PR #12, opened 2024-01-18
Pxweb2 9 install chromatic

SjurSutterudSagen:
Since we want to test out Chromatic in addition to hosting Storybook on netlify, we need to set up automatic deployments to chromatic's systems also.


oyessb (approved):
👍

PR #14, opened 2024-01-22
Added LICENSE.md and SECURITY.md

michaelpande:


likp:
I think taht we should juse Apache License v2. At least we are using that for the old PxWeb code

PR #22, opened 2024-01-31
Added all semantically named colors

michaelpande:
Added all semantically named colors


SjurSutterudSagen (approved):
Can't really comment on what we lose/gain by having the colours directly in this file, instead of a different one with the names the designers use like 'red-400' and importing those. I've never worked with a system like this, or designers, so I'll just trust in your greater experience here :)

LGTM

PR #24, opened 2024-01-31
PXWEB2-39 Run Chromatic action when push to main

SjurSutterudSagen:
Updates the Chromatic Github Action to run on pushes to the main branch, since chromatic needs this for the UI Reviews to work.

Did not update the workflow to trigger on all pushes, even though that might be what chromatic recommends. We do not know if we will need that yet, and we can change it when we see the need for it.


oyessb (approved):
👍

PR #26, opened 2024-02-02
Move storybook-a11y to dev dependencies

SjurSutterudSagen:
Since the addon was installed as a normal dependency by mistake, we move it to dev dependencies. This also updates it from 7.6.10 to 7.6.12.


oyessb (approved):
👍

PR #27, opened 2024-02-05
PXWEB2-61 Added PxWeb default font

MikaelNordberg:
Local font files added for Roboto which will be the default PxWeb font. 
Added twice. For PxWeb app and for Storybook.
Will be possible to change by physically replace them with other font files.
Defined three PxWeb fonts with different font weight.


SjurSutterudSagen (approved):
Looks good

KentMossback (approved):
Ok!

PR #28, opened 2024-02-06
PXWEB2-61 Added folders for typography

oyessb:


SjurSutterudSagen (approved):
Should the sizes for fonts be rem instead of px? Just so we don't use the wrong one and forget to fix it.

Approving since this is just the basic folder structure for the other, new tasks for the components.

SjurSutterudSagen (approved):
All good now :)

PerIngeVaaje (commented):
OK

PR #29, opened 2024-02-06
Use CSS variables instead

michaelpande:


SjurSutterudSagen (approved):
Looks good, this should enable the customization we need after build

PR #30, opened 2024-02-06
Pxweb2 87 ingress

KentMossback:
Added ingress component


libs/pxweb2-ui/src/lib/components/Typography/Ingress/Ingress.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import styles from './Ingress.module.scss';
+
+interface IngressProps {
+  children: React.ReactNode;  
+}
+
+export const Ingress: React.FC<IngressProps> = ({ children }) => {
+  return (
+    <div className={styles.ingress}>
```
SjurSutterudSagen, line 10: Should we return something other than a div here? 'p' maybe?

michaelpande: A <div> is a block. Which means it does not work as text. If you place a ingress inside a <p> tag with another piece of text, it would not be on the same line. The div block would not be on the same line as the rest of the text.

Consider using a span instead. A <p> tag is also a bit limiting, as it could not be used inside another <p>-tag. As that is semantically not valid. 

Alternatively the .ingress style in css could use display: inline, then the div would work as a span.

libs/pxweb2-ui/src/lib/components/Typography/Ingress/Ingress.module.scss
```diff
@@ -0,0 +1,11 @@
+.ingress {
+    font-family: PxWeb-font-400;
+    font-size: 1.25rem;
+    //font-style: normal; // ??
```
michaelpande, line 4: Consider removing the unused code

michaelpande (approved):

apps/pxweb2/public/theme/variables.css
MikaelNordberg, line 1: Should not be part of this PR

MikaelNordberg (commented):
Looks good except for changes in variables.css

MikaelNordberg (dismissed):
Looks good

MikaelNordberg (approved):
Looks good

PerIngeVaaje (approved):
Looks OK

PR #35, opened 2024-02-07
Feature/pxweb 88 body long

PerIngeVaaje:
New try

### DIFF
--- a/libs/pxweb2-ui/src/lib/components/Typography/BodyLong/BodyLong.module.scss
+++ b/libs/pxweb2-ui/src/lib/components/Typography/BodyLong/BodyLong.module.scss
@@ -1,0 +1,29 @@
+@use "../../../../../style-dictionary/dist/scss/fixed-variables.scss" as fixvar;
+
+.bodylong{
+align-self: stretch;
+color: var(--px-color-text-default)
+}
+
+.small{
+        font-family: PxWeb-font-400;
+        font-size: 0.875rem;
+        font-style: normal;
+        font-weight: 400;
+        line-height: 1.5rem; /* 171.429% */
+}
+.medium{
+    font-family: PxWeb-font-400;
+    font-size: 1rem;
+    font-style: normal;
+    font-weight: 400;
+    line-height: 1.75rem; /* 150% */
+}
+
+.small-spacing{
+    margin-bottom:fixvar.$spacing-6
+}
+.medium-spacing{
+    margin-bottom:fixvar.$spacing-7
+}
+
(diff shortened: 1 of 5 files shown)


libs/pxweb2-ui/src/lib/components/Typography/BodyLong/BodyLong.module.scss
```diff
@@ -0,0 +1,29 @@
+@use "../../../../../style-dictionary/dist/scss/fixed-variables.scss" as fixvar;
```
michaelpande, line 1: It looks like the file hasn't been formatted by prettier, do you have the extension installed in vscode?

libs/pxweb2-ui/src/lib/components/Typography/BodyLong/BodyLong.module.scss
```diff
@@ -0,0 +1,29 @@
...
+        font-weight: 400;
+        line-height: 1.5rem; /* 171.429% */
+}
+.medium{
+    font-family: PxWeb-font-400;
+    font-size: 1rem;
+    font-style: normal;
+    font-weight: 400;
+    line-height: 1.75rem; /* 150% */
+}
```
michaelpande, line 21: Several of these are equal, could the ones that are common be moved to .bodylong?

libs/pxweb2-ui/src/lib/components/Typography/BodyLong/BodyLong.tsx
```diff
@@ -0,0 +1,40 @@
...
+  children,
+  size = 'medium',
+  align = 'start',
+  color = 'default',
+  weight = 'regular',
+  spacing = false,
+  ...rest
+}: BodyLongProps) {
+  return (
+    <p
```
michaelpande, line 24: Consider using <span> instead, a <p> tag could be limiting in the future. We can always place the component within a <p> tag if we'd like to use a <p> tag. 

<p><BodyLong>text</BodyLong></p>

The paragraph tag has semantical meaning, and we might not want to use a <p> tag everywhere we want to use bodylong.

michaelpande: The `<p>` tag also adds some margin we might not always want

michaelpande (approved):
Consider using span instead of `<p>`, otherwise LGTM 👍

libs/pxweb2-ui/src/lib/components/Typography/BodyLong/BodyLong.tsx
michaelpande: Discussed on teams, resolved.

PR #38, opened 2024-02-12
Feature/pxweb2 90 label

MikaelNordberg:
Label component


libs/pxweb2-ui/src/lib/components/Typography/Label/Label.tsx
```diff
@@ -0,0 +1,36 @@
...
+export function Label({
+  size = 'medium', 
+  align = 'start',
+  textcolor = 'default',
+  visuallyHidden = false,
+  children,
+  ...rest
+}: LabelProps) {
+  return (
+    <p
```
michaelpande, line 21: This should probably be a span. A `<p>` tag is for paragraphs

MikaelNordberg: @michaelpande 

Now is a label element with the possibility to set htmlFor:

>       <Label htmlFor="fname">First name:</Label><br />
>       <input type="text" id="fname" name="fname" defaultValue="John" /><br />
>

KentMossback (dismissed):
Looks good

KentMossback (approved):
Still looks good

PerIngeVaaje (approved):
Looks OK

PR #44, opened 2024-02-22
Update dependencies with nx migrate latest

SjurSutterudSagen:
Since we want to keep the repository as up-to-date as possible, we need to do the steps required by nx to update it's dependecies.

My current understanding is that you run "nx migrate latest", and then run and apply any migrations create. Then you run "npm update" to update all the dependencies to the current restrictions in the package.json and package-lock.json files.

I do not think this updates any packages not installed by an nx plugin, like the clsx or style-dictionary packages, but as of now they are the latest versions.

Adding the migrations.json file, since the docs advize you to keep it until all branches made before it was created are merged.

Done as part of working on PXWEB2-76.


PerIngeVaaje (approved):
Looks good to my knowledge

PR #54, opened 2024-02-27
Remove an old, unused package-lock file

SjurSutterudSagen:
Since we want the project to be as clean as possible, we should remove old/unused/created-by-mistake files.

This package-lock.json file was created by mistake when we installed style-dictionary the first time. This file should be completely unused and not needed. Nx should build any files pxweb2-ui needs, from the root level package.json and package-lick.json files.


MikaelNordberg (approved):
Looks good

PR #64, opened 2024-03-05
Add typesafety/intellisense to translation keys

SjurSutterudSagen:
Since we are going to be translating the app with i18next, it will be really helpful to get feedback during dev-time for mistakes made with the translation keys. This will give us errors when we try to use keys that do not exist, and also add auto-complete for the translation keys.

The auto-complete is a bit fiddly, and in "t('')" it seems to be activated by adding the ''. So if you don't get it, remove and re-add them.

Also added an example of using the Trans component for i18n.

Lastly, to build the 'resources.d.ts' file from our translation files, I installed the package i18next-resources-for-ts. It is fairly new, but has a score of 75 in the snyk-advisor. One thing to consider here is that, this is only needed since we use JSON files for storing the translations, if instead we used .ts file, this might not be needed. However, our current users are used to JSON files, so did not change this.

I was not sure which pre-script to add this new script to, if any. The file is only needed during development, and is not used during runtime. So neither the prestart or prebuild scripts seem to fit. For now, to get the typesafety, we need to run "npm run build-i18n-interface" when we start developing, or when we update the 'en' translation files.


MikaelNordberg (approved):
Looks good

PR #65, opened 2024-03-07
Feature/pxweb2 99 formatting numbers based on language

SjurSutterudSagen:
Since we want to format numbers based on the language chosen, we need to
add this to the project. The basic functionality was added when i18next
was installed, but we have added examples for how to use this in a few
different ways.

We noticed that a basic example led to many places to set the
roundingMode, and therefore created a function and a custom formatter
for i18next, where we set this roundingMode based on a variable. This
prepares the number formatting for the work to be done later, with a
config file for the project.

Co-authored-by: Mikael Nordberg <54406826+MikaelNordberg@users.noreply.github.com>

### DIFF
--- a/apps/pxweb2/public/locales/en/translation.json
+++ b/apps/pxweb2/public/locales/en/translation.json
@@ -9,13 +9,13 @@
     "simple_date_with_time": "{{value, datetime(year: 'numeric'; month: 'numeric'; day: 'numeric'; hour: 'numeric'; minute: 'numeric')}}"
   },
   "number": {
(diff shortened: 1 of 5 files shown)


SjurSutterudSagen:
The issue we ran into with TypeScript and "roundingMode" missing in the type for intl.NumberFormatOptions is solved https://github.com/microsoft/TypeScript/issues/52072. But this is not yet available in the released ts version, as of now.

michaelpande (approved):
LGTM 👍

PR #68, opened 2024-03-08
Reorganized translation file structure

MikaelNordberg:
The structure of the translation files have been reorganized into sections that follow the structure of the PxWeb user interface.
Hopefully this will make it easy to find the translation text you are looking for.

Only the english translation file is totally updated. The other files only have a minimum of translations.

This is only a draft structure that might come to change later on. It is also possible to divide the translation file into multiple files later on.


michaelpande (approved):
LGTM 👍

PR #74, opened 2024-03-13
Feature/pxweb2 63 add global breakpoint tokens

SjurSutterudSagen:
Since we need to control the layout for different sized screens and such, we need the breakpoints for the media queries. This commit adds the breakpoints to a file, and adds that to the style-dictionary build command. The same as how we define and use the Spacings. In this way we have 1 place where all the breakpoint values are set, and can change them or add to them later if needed.

One thing to consider is that currently the breakpoint tokens are imported (@use) from the files that are generated in the ui lib, and not from inside the main app itself. I am not sure if we want this, or to create another set of files inside the app's own folder, and import from there. This is very easy to add in our current implementation, just add another output file to the build.js in the lib. And update the .gitignore if we don't want them committed to the repo.

Also added an example of the breakpoints working to the main app.

### DIFF
--- a/libs/pxweb2-ui/style-dictionary/README.md
+++ b/libs/pxweb2-ui/style-dictionary/README.md
@@ -5,4 +5,4 @@ Go to the root of the entire project and run
 
 ## Custom theme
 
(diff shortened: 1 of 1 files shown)


MikaelNordberg (approved):
Looks good

MikaelNordberg (approved):
Looks good

MikaelNordberg (approved):
Looks good

PR #75, opened 2024-03-15
Feature/pxweb2 130 data access structure

MikaelNordberg:
PxWeb 2.0 will get data and metadata from PxWebApi 2.0.
The basic infrastructure for the data access has been added:
- An empty typescript library `pxweb-api-client`. This library will contain the code for accessing PxWebApi later on.
- The folder `shared-types` in the `pxweb2-ui` library. This folder will contain shared types that will be used by the UI components.
- The folder `mappers` in the `pxweb2` application. This folder will contain mappers that maps the responses from PxWebApi into the shared types used by PxWeb.

When finished it will look something like this:

```
|- PXWEB2
   |- apps
      |- pxweb2
         |- mappers
            |- ...
            |- table-metadata-response-mapper.ts
            |- ...
   |- libs
      |- pxweb2-ui
         |- shared-types
            |- ...
            |- variable.ts
            |- value.ts
            |- px-table.ts
            |- ...
      |- pxweb2-api-client
```

### DIFF
--- a/libs/pxweb2-api-client/.eslintrc.json
+++ b/libs/pxweb2-api-client/.eslintrc.json
@@ -1,0 +1,30 @@
+{
+  "extends": ["../../.eslintrc.json"],
+  "ignorePatterns": ["!**/*"],
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+      "rules": {}
+    },
+    {
+      "files": ["*.ts", "*.tsx"],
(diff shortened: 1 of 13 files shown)


SjurSutterudSagen (approved):
Looks good to me, but you probably want someone more experienced with structuring projects to also look at it.

Approving it so you can merge it later.

michaelpande (approved):
LGTM 👏 🚀

PR #77, opened 2024-03-18
Feature/pxweb2 130 shared data types

MikaelNordberg:
Shared types that will be used by PxWeb 2.0 and its UI components.
This is only a first version. More properties will be added to the data types later on.


SjurSutterudSagen (approved):
Looks good to me

PR #78, opened 2024-03-18
Feature/pxweb2 149 common text styling

PerIngeVaaje:
Make use of text-styles in common scss

### DIFF
--- a/libs/pxweb2-ui/src/lib/components/Typography/Label/Label.tsx
+++ b/libs/pxweb2-ui/src/lib/components/Typography/Label/Label.tsx
@@ -1,4 +1,5 @@
 import cl from 'clsx';
+
 import classes from './Label.module.scss';
 
 export interface LabelProps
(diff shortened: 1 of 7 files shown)


MikaelNordberg (approved):
Looks good

libs/pxweb2-ui/src/lib/components/Typography/Label/Label.tsx
```diff
@@ -1,31 +1,33 @@
 import cl from 'clsx';
 import classes from './Label.module.scss';
```
SjurSutterudSagen, line 2: Maybe add a blank line between imports from packages and imports from within the project? To better see which imports are what? The same applies to Button.tsx and Tag.tsx too.

```suggestion

import classes from './Label.module.scss';
```

SjurSutterudSagen (dismissed):
Looks good to me! We probably want more feedback on the placement of the text-styles.scss file next to the global.scss file. But it makes sense to me.

SjurSutterudSagen (approved):
LGTM

PR #83, opened 2024-03-19
Feature/pxweb2 94 pxweb2 api client

MikaelNordberg:
Generated api code with openapi-typescript-codegen that was added to the
pxweb2-api-client library. When generating the code the PxAPI-2.yml file
was used as input.

There was a couple of problems with the generated code:
- In the model file for Language.ts both uppercase and lowercase was
used. Edited this file manually.
- The model file for Dataset.ts had a lot of problems. Outcommented the
whole file.

A call to get TableMetadataResponse was added to the application.
Variable names are listed for the table. Possible to select language for
the API calls.

API calls are made to https://pxapi2-master-px.scb.se/api/v2 which only
has support for english and swedish.


KentMossback (approved):
A lot of code but looks good to me

PR #84, opened 2024-03-21
Feature/pxweb2 131 convert api response

MikaelNordberg:
The response from the API is mapped to PxWeb data types that is then used by the UI components in PxWeb


PerIngeVaaje (approved):
Looks good to me after a good review from developer

likp (approved):

PR #85, opened 2024-03-22
Feature/pxweb2 33 search new

KentMossback:
The search component


KentMossback:
Rebase in progress

SjurSutterudSagen:
libs/pxweb2-ui/src/lib/components/Search/Search.tsx,
libs/pxweb2-ui/src/lib/components/Search/Search.module.scss and 
libs/pxweb2-ui/src/lib/components/Search/Search.stories.tsx
Has some different indentations, could you format the files to fix it? 

Also, in libs/pxweb2-ui/src/lib/components/Search/Search.module.scss, `padding: var(--spacing-2, 8px);` only works because of the 8px fallback. Maybe it would be better to use `gap: fixed.$spacing-2;` since you already import those globals at the top of the css file.

Per Inge also noticed, in the css, that you set the typography on the .input class, but that should be from the global style called `/* BodyShort/Medium */` according to figma, so can you import that one and use that instead?

libs/pxweb2-ui/src/lib/components/Search/Search.tsx The imports are now mixed between external packages and internal files. Could you move the React imports to top, before or after the clsx one, and then add a blank line? I think that makes it cleaner and easier to see what is what in the imports.

Lastly, could you google translate the Arabic version, and add that to it's translation file?

KentMossback:
Merged code from main. Made adjustments suggested in comment from @SjurSutterudSagen

SjurSutterudSagen (approved):
LGTM

PR #93, opened 2024-04-02
Fix the variable name for pxTable state

SjurSutterudSagen:
Since we want to follow best practices as early as possible, we should use the best practices nameing schemes for state variables, even in the early test phase.


MikaelNordberg (approved):
Looks good

PR #94, opened 2024-04-02
Update dependencies

SjurSutterudSagen:
Since we want up-to-date dependencies, we need to regularly update them.

Ran the nx migrate command, npm update --save and ran npm ci to test that it still builds.


MikaelNordberg (approved):
Looks good

PR #95, opened 2024-04-03
Delete obselete deploy.yml for Github Pages deployment

SjurSutterudSagen:
Since we no longer want to deploy to Github Pages, we should remove the old deploy.yml for that Github Action.


michaelpande (approved):
LGTM 👍

PR #96, opened 2024-04-03
Patch/pxweb2 154 update chromatic to v11

SjurSutterudSagen:
Since we want to keep the project up-yo-date, we need to update packages
that have breaking changes. The Chromatic v11+ breaking change has to do
with TurboSnap, and a new check for setting/telling the cli the baseDir
for storybook.

Since we do not use TurboSnap, and only used 1500 of our 5000 free
snapshots last month. Adding this can be done later, when we need to
look at ways to optimize the build process. Or we can drop it
entirely if we can upgrade our Chromatic project account to an Open
Source one. Though we still might want to do it just to not be wasteful.

### DIFF
--- a/libs/pxweb2-ui/style-dictionary/README.md
+++ b/libs/pxweb2-ui/style-dictionary/README.md
@@ -5,4 +5,4 @@ Go to the root of the entire project and run
 
 ## Custom theme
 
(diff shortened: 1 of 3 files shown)


michaelpande (approved):
LGTM 👍

PR #97, opened 2024-04-04
Fix warnings cause by argument names in Icons list

SjurSutterudSagen:
Since we don't want warnings in our application, we should fix the ones we do have. We received warnings when running and building the component library, about wrong arguments used in the Icons list:

`Invalid DOM property "fill-rule"`
`Invalid DOM property "clip-rule"`


PerIngeVaaje (approved):
Looks ok

PR #98, opened 2024-04-08
Feature/pxweb2 156 add text style to heading

SjurSutterudSagen:
Since we want to use global text styles, instead of each component
having it's own text styling, we need to update the Heading component.

This also updates the text styles of the Heading component to the
latest ones from Figma.


MikaelNordberg (approved):
Looks good

PR #103, opened 2024-04-09
Patch/update dependencies

SjurSutterudSagen:
Since we want the project to be up-to-date, we need to update the npm packages in the project.

This updates both the Nx packages themselves, and the other packages through 'npm update --save'.

This PR also will close the 4 "Bump" PRs  from dependabot that are currently open in the project.


MikaelNordberg (approved):
Looks good

PR #107, opened 2024-04-10
Update dependencies

SjurSutterudSagen:
Since we want to keep the project dependencies up-to-date, we need to update them when they are outdated. This updates Nx to the newest patch version, and the other dependencies to the newest versions allowed from that.


MikaelNordberg (approved):
Looks good

PR #109, opened 2024-04-11
Feature/pxweb2 52 select component take3

MikaelNordberg:
Select component.
Possible to add css classes to the Button and BodyShort components from the code using them.
Use common text styles in BodyShort.

### DIFF
--- a/libs/pxweb2-ui/src/lib/components/Select/Select.module.scss
+++ b/libs/pxweb2-ui/src/lib/components/Select/Select.module.scss
@@ -1,10 +1,3 @@
-/*
- * Replace this with your own classes
- *
- * e.g.
- * .container {
- * }
-*/
 @use '../../../../style-dictionary/dist/scss/fixed-variables.scss' as fixed;
 
 // ----------------------------------------------------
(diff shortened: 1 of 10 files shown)


libs/pxweb2-ui/src/lib/components/Select/Select.tsx
```diff
@@ -0,0 +1,143 @@
...
+  label: string;
+  value: string;
+};
+
+export interface SelectProps {
+  variant?: 'default' | 'inVariableBox';
+  label: string;
+  hideLabel?: boolean;
+  defaultOption?: string;
+  options: SelectOption[];
```
michaelpande, line 17: We should probably add onChange here. The purpose of passing a onChange function to the select component is to send data out of the component when a new set of items is selected. A example of how this is done can be seen in the checkbox component 😄 

When selecting options in the modal component the data will then flow from the modal to this component, and then have to flow out of this component to the parent. onChange is how we solve this.

michaelpande (changes requested):

libs/pxweb2-ui/src/lib/components/Select/Select.module.scss
```diff
@@ -0,0 +1,150 @@
...
+    border: 0 !important;
+    clip: rect(0, 0, 0, 0) !important;
+    height: 1px !important;
+    margin: -1px !important;
+    overflow: hidden !important;
+    padding: 0 !important;
+    top: 0 !important;
+    position: absolute !important;
+    white-space: nowrap !important;
+    width: 1px !important;
```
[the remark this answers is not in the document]
MikaelNordberg: Gave the same result with less code. I will change that.

libs/pxweb2-ui/src/lib/components/Select/Select.tsx
MikaelNordberg: Now I have added the onChange event

michaelpande (dismissed):
LGTM 👍

libs/pxweb2-ui/src/lib/components/Select/Select.tsx
```diff
@@ -0,0 +1,164 @@
+import cl from 'clsx';
+import classes from './Select.module.scss';
```
SjurSutterudSagen, line 2: Can we add a blank space between imports of the external packages and the internal files? Makes it easier to parse them for me atleast. So between the current lines 1 and 2.

PR #116, opened 2024-04-17
Update dependencies

SjurSutterudSagen:
Since we want to keep the dependencies in the project up-to-date, we need to update the outdated ones. This updates them with the 'migrate latest' command from Nx, and then with the normal 'update --save' from NPM.

Also add a blank space to a readme in the ui lib, to force chromatic to build, to check for regressions.

This should also close the current "Bump" PRs from dependabot.


PerIngeVaaje (approved):
Looks  good to me

PR #117, opened 2024-04-17
Pxweb2 100 config

michaelpande:
- Added config that can be changed before and after build. 
- Added config validation with JSON schema, validated on localhost with ajv (this supports validation for config changes after build as well)
- Added config types
- Used config in language config 
- Made sure config is loaded before app
- Config is available anywhere in the app via getConfig()


michaelpande:
**Discussion**: I considered adding the config to index.html in the following manner to prevent an additional request to the server. I decided to go with a seperate file as it seemed like a better approach, but i'd like feedback on it. 
image

SjurSutterudSagen (approved):
LGTM

PR #119, opened 2024-04-18
Refactor Select into 2 sub-components

SjurSutterudSagen:
To get rid of an extra div that the Select component returns, it has been refactored into being 2 sub components for the variants, instead of 2 functions. Nothing has been done to the logic outside of this.


MikaelNordberg (approved):
Looks good

PR #127, opened 2024-04-24
Feature/pxweb2 115 modal

MikaelNordberg:
Modal component added and used in the Select component


SjurSutterudSagen (approved):
LGTM, let's get it in so we can use it in the VariableBox

PR #128, opened 2024-04-25
Update ci-tests.yml

SjurSutterudSagen:
Add a name to the job that runs the tests, so the workflow/job (hopefully) show up in the branch protection rules lists, so we can enforce passing the tests in PRs.


PerIngeVaaje (approved):
OK

PR #129, opened 2024-04-25
Make tests work

MikaelNordberg:


SjurSutterudSagen (approved):
Let's see how the PR checks behave with a failing test that is changed

SjurSutterudSagen:
The tests are the only checks that have failed, and the PR can't be merged. Working as intended then :)

SjurSutterudSagen (approved):
Approving removing the test for modal here, since we just need the main branch's ci run to finish without failing on the tests.

PR #138, opened 2024-05-07
Feature/pxweb2 55 add variable box component

SjurSutterudSagen:
Add the current work on the VariableBox component into the development main branch. 

Currently it looks and behaves in large parts like we want, but is missing some features and logic. State is also currently almost completely internal, and some of it needs to be hoisted up to the parent component.

Implemented logic:
VB Header is done
 - Can open/close
 - Shows correct language
 - Shows correct information
VB Content is not done
 - Shows correct components, except it is missing Alert (not finished yet), Radio buttons (not finished), Metadata link (not implemented in the API, so on hold for now)
 - Missing logic for updating the values shown in the list when selecting a codelist in the select
 - Focus marking, is it's own task in the backlog
 - handling search,  is it's own task in the backlog

Check the Jira-board for a complete list of what is missing for the entire VB to be implemented fully.


MikaelNordberg (dismissed):
Looks good for now.

MikaelNordberg (approved):
Looks good for now

PR #145, opened 2024-05-15
Since we want to keep the dependencies in the project up-to-date, we

PerIngeVaaje:
need to update the outdated ones. This updates them with the 'migrate latest' command from Nx, and then with the normal 'update --save' from NPM.

Also add a blank space to a readme in the ui lib, to force chromatic to build, to check for regressions.


SjurSutterudSagen (approved):
Looks good, lets keep it up-to-date

PR #157, opened 2024-05-29
Update the project through nx migrate latest

SjurSutterudSagen:
Since we want to keep the dependencies of the project up-to-date, we need to update them regularly. This updated the dependencies with 'nx migrate latest" and then "npm update --save".


MikaelNordberg (approved):
Looks good

PR #159, opened 2024-05-30
Update project dependencies

SjurSutterudSagen:
Since we want to make sure the project dependencies are not outdated, we need to update them. This is in addition to the update deps PR from yesterday, since Sass had an update right after that PR was merged in.

It also seems they added Licence information to the package-lock file, so the changes in there seems larger than they are.


michaelpande (approved):
👏

PR #160, opened 2024-05-30
Fix missing cursor pointer on VariabelBoxHeader

SjurSutterudSagen:
Since we want the indications for when something can be pressed to be clear, we need to change the mouse cursor when hovering over the clickable header of the VariableBox component.


MikaelNordberg (approved):
Works fine!

PR #162, opened 2024-06-03
Fix using wrong units for setting font-values

SjurSutterudSagen:
Since we want the user's font size settings in the browser to be respected, we need to use units for setting the font-sizing and similar, that respect those settings. Since "px" does not allow this, we need to change the places where they are used for that.

The visual problems with showing multiple lines in checkboxes are not because of this change. That is because of "min-height" on Checkbox, and "max-height" on it's container do not work correctly together. It will be fixed separatly.


MikaelNordberg (approved):
Looks good

PR #166, opened 2024-06-05
Update dependencies with Nx migrate latest

SjurSutterudSagen:
Since we want to keep the project updated, we need to update our outdated dependencies. This has updated the dependencies with Nx migrate latest and npm update --save.

It should remove the one "bump" PR we currently have, but it does not update dependencies that are not linked to Nx, and have updates available but that are outside the allowed range in package.json.


michaelpande (approved):
LGTM 👍

PR #167, opened 2024-06-05
Capitalize the first letter of valuename in VB

SjurSutterudSagen:
Since we want the language in the application to look correct and professional, we need to make sure we capitalize the first letter of value names in the list of checkboxes in the variable box.

Line 17 was changed by the autoformatter.


PerIngeVaaje (approved):
OK

PR #171, opened 2024-06-07
Fix/hover on search in variablebox

SjurSutterudSagen:
When hovering, or focusing the Search in the VariableBox, the outline was only showing in the corners of the Search.

This turned out to be two problems, the bottom of the outline was going behind the background of the MixedCheckbox. Which only needed some z-index adjusting to work correctly. 

The other problem is that the outline goes outside the scrollable list, but with `overflow-y: auto` these are then removed therefore. I could not figure out how to solve this completely, without having to use alot more time. I considered 2 solutions here:

1. To moved the search outside the scrollable list. This should make it look correct, but would remove the scrolling functionality of it.
2. To add an offset on the outline, so it is inside of the scrollable list.

I chose to implement option 2 since it was 1 line of code, while keeping the scrolling intact. But this will be discussed with the designer on Monday, and is a TEMPORARY fix.

Also removed defaults in the CSS, since we want to be certain where CSS values come from.

### DIFF
--- a/libs/pxweb2-ui/src/lib/components/Search/Search.module.scss
+++ b/libs/pxweb2-ui/src/lib/components/Search/Search.module.scss
@@ -40,6 +40,7 @@
   border-radius: var(--px-border-radius-medium);
   background: var(--px-color-surface-subtle);
   z-index: 1;
+  outline-offset: -2px;
 
   &:hover {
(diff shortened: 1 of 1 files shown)


MikaelNordberg (approved):
Looks good

PR #180, opened 2024-06-10
PxWeb2-209: Fix keyboard navigation

SjurSutterudSagen:
Since many of our user use a keyboard to navigate on the site, the VariableBox component needs to handle this correctly. This adds the navigation order according to the Figma spec. The arrow down key is used to enter into the list of checkboxes for the variables, and moves the focus to the first checkbox in the list. The arrow up key is then used to leave it again, which moves the focus back to the list.

One thing that isn't handled, is when a user moves into a list with the arrow down key, clicks out with the mouse, and then moves back to that list with the keyboard. It will then go into that list, instead of to the next variable. I deemed this to be something we can fix later, since you should not ever have this problem with only keyboard navigation.

Also, since we want keyboard users to under stand how to use the different parts of the VariableBox, we need to inform them of what the list of values is, and how to use it.

The aria-description is coming fully in WAI-ARIA 1.3, but browser support is at 88% according to caniuse.com. Should we change this to be in the aria-label instead? It feels too long to be the right choice here.

Or if there is a better way to tell the users about navigating into and out of the values list, we can change it.


PerIngeVaaje (approved):
OK

PR #182, opened 2024-06-12
Feature/px web2 231 add alert to variablebox

SjurSutterudSagen:
Since we want to show when there are no selected values for a mandatory
variable, we need to add this capability to the VariableBox component.

When no value is selected, it now shows and alert in the header of the
component, and the border is colored red. This is removed when a value
is chosen.

API handling is not added, that would be a different task.

Also had to add a new folder to ignore files. This is for the newest changes 
in Nx, and those lines are also added in the PR for those updates.

### DIFF
--- a/libs/pxweb2-ui/src/lib/components/VariableBox/VariableBox.tsx
+++ b/libs/pxweb2-ui/src/lib/components/VariableBox/VariableBox.tsx
@@ -52,6 +52,9 @@ export function VariableBox({
   const chosenValuesLength = selectedValues.find(
     (variables) => variables.id === id
   )?.values.length;
+  const currentVariable = selectedValues.find((variable) => variable.id === id);
+  const hasMandatoryNoSelectedValuesError =
+    mandatory && (currentVariable?.values.length === 0 || !currentVariable);
 
   if (prevTableId !== tableId) {
     setIsOpen(false);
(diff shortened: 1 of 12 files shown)


SjurSutterudSagen:
Checking to see if the animations in the design will take alot of time or not. If not I'll add them too.

libs/pxweb2-ui/src/lib/components/VariableBox/VariableBox.tsx
```diff
@@ -52,6 +52,9 @@ export function VariableBox({
   const chosenValuesLength = selectedValues.find(
     (variables) => variables.id === id
   )?.values.length;
+  const currentVariable = selectedValues.find((variable) => variable.id === id);
+  const hasMandatoryNoSelectedValuesError =
```
michaelpande, line 56: Consider using another variable name. This one is a bit difficult to follow. Maybe isMissingMandatoryValueError?

michaelpande (approved):
Nothing big, just some minor comments. Looks good 👍

libs/pxweb2-ui/src/lib/components/VariableBox/VariableBoxHeader/VariableBoxHeader.tsx
```diff
@@ -52,40 +55,56 @@ export function VariableBoxHeader({
...
         </div>
       </div>
 
-      <div className={cssClasses}>
-        {isOpen ? (
-          <Icon iconName="ChevronUp"></Icon>
-        ) : (
-          <Icon iconName="ChevronDown"></Icon>
+      <div className={cl(classes['header-alert'])}>
+        {isMissingMandatoryValues && (
```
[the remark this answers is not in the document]
SjurSutterudSagen: Yes, I'll fix it :)

libs/pxweb2-ui/src/lib/components/VariableBox/VariableBox.tsx
SjurSutterudSagen: Changed to use this name

PerIngeVaaje (approved):
OK

PR #183, opened 2024-06-13
Fix h-tag mistakenly used in VarBox header element

SjurSutterudSagen:
The header of the variablebox should not have an actual h-tag.

Change it to a p-tag with the correct styling.


MikaelNordberg (approved):
Looks good

PR #186, opened 2024-06-14
Feature/pxweb2 204 map jsonstat2 data

MikaelNordberg:
The Table component now uses data and metadata from json-stat2 response from the API


SjurSutterudSagen:
Should we have some tests for something? Otherwise looks good to me.

SjurSutterudSagen (approved):
LGTM now, good job!

PR #187, opened 2024-06-14
Add default selection to the VariableBox

SjurSutterudSagen:
Since we need to show the selections made in the default selection algorithm, we need to add this to the logic for the table view/app.tsx. This is a first version to get it working, with a basic test for the mapper created to map the response from the API. It also updated the variable box when a new language is chosen.

I did find that the current url created is too long in the large table we test on, therefore I have only tested this one the three smaller test tables in the app.

The blinking error state that appears is part of the handling of a loading state in the app page, I think.


SjurSutterudSagen:
Wierd, the chromatic actions don't progress... Debugging

MikaelNordberg (approved):
Looks good

PR #188, opened 2024-06-17
Set cursor to pointer for Button-component

PerIngeVaaje:
Added cursor to pointer for Button  component


SjurSutterudSagen (approved):
LGTM

PR #192, opened 2024-06-18
Feature/px web2 196 error in variablebox

SjurSutterudSagen:
Since we want the users to be able to tell when there is an
error/warning or similar, we need to clearly show this when one occurs.

This adds the first version of the EmptyState component, that is
intented to be used for this, however, there are 2 other tasks to fully
implement this in the backlog. This is just a temporary, first version,
which only contains text for missing mandatory values. And the logic to
show it when that happens, instead of the table.


PerIngeVaaje (approved):
OK

PR #193, opened 2024-06-19
Fix/pxweb2-266 Fix red error when loading metadata

SjurSutterudSagen:
Since we do not want the users to see a red error whenever we load metadata for mandatory variables, we need to make sure we block the error when metadata is loading.

This needs to be looked at again, when we look into proper loading states for the page, since this might be not needed when that is implemented.


michaelpande (approved):
LGTM 👍

PR #194, opened 2024-06-19
Added routing & updated logos to svg

michaelpande:


apps/pxweb2/src/app/app.tsx
```diff
@@ -233,14 +237,20 @@ export function App() {
...
+    if (!tableId) {
+      return;
+    }
+    TableService.getMetadataById(selectedTableId, i18n.resolvedLanguage)
+
     if (isLoadingMetadata === false) {
       setIsLoadingMetadata(true);
     }
 
     TableService.getMetadataById(tableid, i18n.resolvedLanguage)
```
SjurSutterudSagen, line 252: This should now be "tableId". Fails here in Cloudflare according to the logs.

SjurSutterudSagen (changes requested):
Fails when building in Cloudflare, seems to be a renamed variablename that was missed.

SjurSutterudSagen:
Currently, we have a problem with how the checks on PRs are enforced. It requires the Chromatic checks, even on PRs that don't need to build chromatic. A workaround until we fix the issue is to add/remove a blank line/space in a Readme.md file inside the "libs/pxweb2-ui/**" folder.

SjurSutterudSagen (approved):
LGTM

SjurSutterudSagen (approved):
LGTM

PR #195, opened 2024-06-19
Fix translation for EmptyState

SjurSutterudSagen:
Now should be the same as the language document in Confluence


MikaelNordberg (approved):
Looks good

PR #200, opened 2024-06-20
Patch/update dependencies

SjurSutterudSagen:
Since we want to keep the project up-to-date, we need to update the
packages. This update the project packages with 'nx migrate' and then
'npm update --save'.

Individual packages not part of the Nx system are not updated.


MikaelNordberg (approved):
Looks good

PR #201, opened 2024-06-21
Fix missing max-width on footer

SjurSutterudSagen:
Simple fix for the Footer component missing the same max-width as the Content component


PerIngeVaaje (approved):
OK

PR #202, opened 2024-06-21
Fix empty state to be at the correct position

SjurSutterudSagen:
Since we only want the EmptyState warning to be rendered at the correct position on the table view, we need to move where the check is being done. This correctly only replaces the table in the table view, and not the entire content view.


PerIngeVaaje (approved):
OK

PR #212, opened 2024-06-25
Update the project with Nx migrate latest

SjurSutterudSagen:
Since we want to keep the project up-to-date, we need to update our dependencies. This updates the Nx packages with the 'nx migrate latest' command, and then with 'npm update'.

The individual packages are not updated here.


MikaelNordberg (approved):
Looks good

PR #213, opened 2024-06-26
Test site for PxWeb 2 with SSB:s API *DO NOT MERGE* 

likp:


SjurSutterudSagen:
Something that could be interesting here to test, is to also change the theme/colors of the test site. So we can check that it is working as intended.

PR #216, opened 2024-06-28
Feature/pxweb2 284 Time periods sometimes in the wrong order in the table

MikaelNordberg:
This PR addresses the following issues:
- Time periods sometimes sorted in the wrong order in the table in Pxweb2 (#284). 
- 3 calls was made to get data from the API (#282)

If selectedVBValues was not in sync with variables context, a data fetch without selected variables and values (default selection) was made. In this call time periods was sorted descending by the API.

The fix assures that selectedVBValues are always in sync with variables context and also a call to fetch data from the API is only made when variables context is updated. In this way a call to get data from the api always contains information about selected variables and values.

The fix also adresses a problem we had when switching tables. Before the selected values for table 1 was used when fetching data for table 2. This caused a http 400 error - Non-existent variable. We were trying to get values that existed in table 1 but not existed in table 2. 
The problem was that the useEffect was triggered by both tableId, selectedVBValues and variables. Now we have limited the useEffect to only be triggered when the variables context (that is in sync with selectedVBValues) or language is changed.

Initially we was about to fix issue #PXWEB2 283 but ended up fixing #PXWEB2 284 and #PXWEB2 282 instead. Therefore the name of the branch do not match the name of the PR. 

Also #PXWEB2 282 is addressed in this fix. The earlier solution was making 3 calls for fetching data from the API (1 for the default selection (because selectedVBValues and the variables context was not in sync) and 2 more calls containing variables and values). With this PR only 1 call is made to fetch data from the API.


apps/pxweb2/src/app/app.tsx
```diff
@@ -238,13 +248,13 @@ export function App() {
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [i18n, selectedVBValues, tableId, variables, i18n.resolvedLanguage]);
+  }, [variables, i18n.resolvedLanguage]); // Should only run this useEffect when selectedVBValues are in sync with variables context
 
   useEffect(() => {
     if (!tableId) {
       return;
     }
-    TableService.getMetadataById(selectedTableId, i18n.resolvedLanguage);
+    //TableService.getMetadataById(selectedTableId, i18n.resolvedLanguage);
```
michaelpande, line 257: Consider deleting it instead of leaving it commented out

apps/pxweb2/src/app/app.tsx
```diff
@@ -273,7 +283,8 @@ export function App() {
                 variable.selectedCodeList !== undefined
             );
 
-            setSelectedVBValues(defaultSelection);
+            updateAndSyncVBValues(defaultSelection);
+            //setSelectedVBValues(defaultSelection);
```
michaelpande, line 287: Consider deleting instead of commenting out

michaelpande: Same goes for everywhere this is done :)

apps/pxweb2/src/app/app.tsx
```diff
@@ -431,12 +447,14 @@ export function App() {
...
 
-      setSelectedVBValues(newSelectedValues);
+      updateAndSyncVBValues(newSelectedValues);
+      //setSelectedVBValues(newSelectedValues);
     }
   };
 
   function handleVBReset() {
     if (selectedVBValues.length > 0) {
+      //updateAndSyncVBValues([]) - Do not sync with context when resetting
```
michaelpande, line 457: Consider deleting this comment

michaelpande (dismissed):
LGTM 👍 I haven't tested it, just looked at the code.

SjurSutterudSagen (approved):
LGTM

PR #217, opened 2024-07-29
Update project dependencies with Nx Migrate

SjurSutterudSagen:
Since we want to keep the project updated, we need to update the dependencies regularly. This updates the project dependencies with the Nx Migrate command. It does not update the non-nx dependencies


PerIngeVaaje (approved):
Looks ok

PR #218, opened 2024-07-29
Update chromatic-cli to latest

SjurSutterudSagen:
Since we want to keep the project updated, we need to update the chromatic cli package to the latest version.

This will be tested when the pipeline runs for the PR.


PerIngeVaaje (approved):
Looks ok

PR #219, opened 2024-07-29
fix/Refactor scss to have nesting at the end of class

SjurSutterudSagen:
Since we do not want our build output to have unnecessary warnings, we need to fix what we can when we have them. This fixes some warnings about nesting in .scss where there are more styling afterwards. This will be changed in a future version of scss.


PerIngeVaaje (approved):
OK

PR #220, opened 2024-07-29
Migrate style-dictionary to new major version: 4.0.1

SjurSutterudSagen:
Since we want to keep up with major versions of the packages we use, we need to update style-dictionary to v4. We were on v3. We have a pretty light setup, so the process was fairly minor. The main changes are that the lib where it is installed needed to use ES Modules, and that parts of the code is now async.

One change to the output is that it now outputs capital letters for colors, so the variables.css file has many changes.

They also changed the auto-generated comment at the top of that file to not have a timestamp, so it should now not show it has changed everytime the app is run.


PerIngeVaaje (approved):
OK

PR #221, opened 2024-07-30
Update react-i18next to v15

SjurSutterudSagen:
Since we want to keep the packages updated, we need to update them when new major versions are released. It seemed that they removed some older, deprecated namespace loading, which is the breaking change. But the changelog is not clear, and this is based on comments made in the issues in the changelog. I could not find any information about what, if any, breaking changes there are for users updating from v14 to v15.

Tested locally, and seemed to work as expected. Will also test with the preview that is deployed in the pipeline.


PerIngeVaaje (approved):
OK

PR #222, opened 2024-07-30
Add check for empty errorMsg to tableprovider

SjurSutterudSagen:
Since we do not want unecessary warnings or errors in the console, we need to remove the ones that appear. This fixes two errors when loading the site, because the useEffect in tableprovider for showing the errors in the console was missing a check for an empty string.


KentMossback (approved):
Looks good!

PR #223, opened 2024-07-30
Fix missing unique keys in variableBox component

SjurSutterudSagen:
Since we do not want unnecessary errors and warnings in the console, we need to fix the long standing issues of the warnings for missing keys for the variableBox component. There were two such warnings, and this should fix both(finally).


KentMossback (approved):
Looks good, no more warnings

PR #224, opened 2024-07-30
Fix/pxweb2 283 remove resetting selection on lang change

SjurSutterudSagen:
Since we no longer want the selection to reset to the default selection
when you change language, we need to change so we retain the selections
made. This removes that reset, and we now only make a call for the
defaultSelection once.


KentMossback (approved):
Looks good and work as it should

PR #226, opened 2024-07-31
fix/PxWeb2-270: Add new Chromatic workflow to skip if no ui change

SjurSutterudSagen:
Since we want to require Chromatic checks to pass before merge to main, we need a workflow to pass those required checks when there are no changes that are related to the UI. As far as I know, there are no ways to test workflows locally before atleast a minimal version is added to main.

The skip workflow is simple, and this is an 'inverted' version of our old workflow. Will need to see if it works in main. Future work and debugging could possibly be done in a new branch, once this file has been added to main so it shows up in GitHub Actions page.

### DIFF
--- a/.github/workflows/chromatic-skip.yml
+++ b/.github/workflows/chromatic-skip.yml
@@ -1,0 +1,33 @@
+# .github/workflows/chromatic-skip.yml
+
+name: 'Chromatic'
+
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


SjurSutterudSagen:
Oh, looks like it worked without the workflow file being merged to main first. Though, maybe we don't need the "Storybook Publish" as required?

KentMossback (approved):
Looks good

PR #228, opened 2024-08-01
patch/Update nx dependencies

SjurSutterudSagen:
Since we want to keep the project dependencies up-to-date, we need to update them. This updated the Nx dependencies to the latest versions.


KentMossback (approved):
Looks good!

PR #230, opened 2024-08-01
fix/PxWeb2-283: Fix for getting defaultSelection for new table

SjurSutterudSagen:
Since we want to only get the default selection when you change tables, we need to correctly keep track of the old table id, so we know when to grab the default selection again.

This is a fix for the prior PR that was merged on this issue, PxWeb2-283.


PerIngeVaaje (approved):
It works as intended

PR #232, opened 2024-08-07
patch/Update nx dependencies

SjurSutterudSagen:
Since we want to keep the project up-to-date, we need to update the dependencies regularly. This updates the Nx dependencies to the newest versions.

It does not update non-Nx dependencies beyond the ranges allowed in package.json


SjurSutterudSagen:
This PR will solve the Dependabot one for bumping react-router-dom

MikaelNordberg (approved):
Looks good

PR #233, opened 2024-08-07
fix/clean up app.tsx

SjurSutterudSagen:
Since we want to refactor the app.tsx file, I thought it would be a good idea to clean it up at bit before hand. This fixes various issues we have with it, like unused imports, unused (old) state etc.


MikaelNordberg (approved):
Looks good

PR #234, opened 2024-08-07
Add a custom theme to nor test-site

SjurSutterudSagen:
Since we want to test the theme system also, we should have a test-site that uses the custom-theme functionality. This adds some very basic color changes to the test-site for the norwegian api.

The color changes are not supposed to be perfect in this commit, this is to show how it looks, and to check that it works in the pipeline.


PerIngeVaaje (approved):
OK!

PR #235, opened 2024-08-08
PXWEB2-216 Added Spinner

PerIngeVaaje:
Added Spinner component. Also corrected storybook path for Alert

### DIFF
--- a/libs/pxweb2-ui/src/lib/components/Spinner/Spinner.tsx
+++ b/libs/pxweb2-ui/src/lib/components/Spinner/Spinner.tsx
@@ -12,25 +12,9 @@ export function Spinner({
   variant = 'default',
   label,
 }: SpinnerProps) {
-  let PathStrokeSubtle = '';
-  let PathStrokeAction = '';
   let Eclipse = null;
   let Labelsize: 'small' | 'medium' = 'medium';
 
-  switch (variant) {
-    case 'default':
-      PathStrokeSubtle = 'var(--px-color-border-subtle)';
-      PathStrokeAction = 'var(--px-color-border-action)';
-      break;
-    case 'inverted':
-      PathStrokeSubtle = 'var(--px-color-border-subtle)';
-      PathStrokeAction = 'var(--px-color-border-on-inverted)';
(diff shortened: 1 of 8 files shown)


libs/pxweb2-ui/src/lib/components/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,244 @@
...
+  size = 'medium',
+  variant = 'default',
+  label,
+}: SpinnerProps) {
+  let PathStrokeSubtle = '';
+  let PathStrokeAction = '';
+  let Eclipse = null;
+  let Labelsize: 'small' | 'medium' = 'medium';
+
+  switch (variant) {
```
SjurSutterudSagen, line 20: This switch and the variables on line 15 and 16 seem to be ununsed. Please remove them.

libs/pxweb2-ui/src/lib/components/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,244 @@
...
+    case 'inverted':
+      PathStrokeSubtle = 'var(--px-color-border-subtle)';
+      PathStrokeAction = 'var(--px-color-border-on-inverted)';
+      break;
+    default:
+      PathStrokeSubtle = 'var(--px-color-border-subtle)';
+      PathStrokeAction = 'var(--px-color-border-action)';
+      break;
+  }
+  switch (size) {
```
SjurSutterudSagen, line 34: Theres a lot of code being repeated in the SVGs here. I'd prefer only changing some variables and then having one svg part, but I'll leave the decision here to you :)

SjurSutterudSagen (changes requested):

libs/pxweb2-ui/src/lib/components/Spinner/Spinner.tsx
PerIngeVaaje: Changes done.

SjurSutterudSagen (approved):
LGTM

PR #236, opened 2024-08-09
TableDataProvider stores a PxTable object instead of a string

MikaelNordberg:
TableDataProvider now stores a PxTable object instead of a string in context.
In this way we do not need to call JSON.stringify and JSON.parse when storing the PxTable in context and using it in our components which is more efficient.


SjurSutterudSagen (approved):
LGTM

PR #239, opened 2024-08-13
Patch/update nx dependencies

SjurSutterudSagen:
Update the Nx dependencies to newest, and the other dependencies to the
newest allowed in package.json.

Does not update dependencies outside of allowed ranges.


PerIngeVaaje (approved):
OK

PR #240, opened 2024-08-14
Feature/pxweb2-207: Add illustration support to empty state

SjurSutterudSagen:
Since we want to have illustrations to help with informing about the EmptyState, we need to implement the functionality for displaying SVGs in the EmptyState component. This expands the EmptyState component to have this functionality, and the component also checks for the screen width to display a small or large illustration. The colors in the illustrations are taken from the CSS colors, but after discussing this with the designer, there might be changes later to the names of these colors. Now we have color names in them, like "bluegray", but this would look weird when you overwrite it with another color. Might end up with a green "bluegray" color.

This also updates the Style-Dictionary build.js file to also build the types for typescript, and the css colors to js.

Since the Content-Container is not done yet, it might not look exactly like figma on small screens.


SjurSutterudSagen:
Waiting for the fix for the CSS reset.

PerIngeVaaje (approved):
OK

PerIngeVaaje (approved):
OK

PR #243, opened 2024-08-16
Fixes for performance  bottlenecks in PxWeb 2

MikaelNordberg:
Fixes:
- Always add values in bulk to state. They should never be added one and one for performance reasons


SjurSutterudSagen (approved):
Looks like a good optimization. Approved!

PR #250, opened 2024-08-21
Fix/refactor rendering of variablebox in app

SjurSutterudSagen:
Since we want to refactor the app.tsx, one of the things we can improve is how we use the VariableBox component. We mapped over the variables and created the VariableBoxes directly in the place it was used. This adds a VariableBoxList component that abstracts that away.

There is still more work to be done here, with minimizing prop-drilling. But that is not part of this PR.

Also add some tests and test data. This might be moved/expanded on later when we improve the tests for the VariableBox


MikaelNordberg (approved):
Good fix!

PR #252, opened 2024-08-21
Fix the global styling for the pxweb2-ui library

SjurSutterudSagen:
Since we are resetting some of the CSS in the app, we want the stories in storybook to reflect the same reset. Currently the reset is only done in the app, which means the components look different between the app and the library.

This change might result in many changes in the stories in the lib, but should not result in any changes in the app.


SjurSutterudSagen:
Fixed the changes that appeared in Chromatic, and approved the change to EmptyState height, which was expected. The EmptyState component will be properly approved in it's own PR, when illustration support is added.

michaelpande:
I don't have time to check it right now, but I have some opinions on this. So wait with the merge, please :)

SjurSutterudSagen:
@michaelpande I fixed the old table styling that was in the global scopes, and it is now moved into the Table component.

PR #256, opened 2024-08-27
Feature/pxweb2 213

KentMossback:
Show scrolls when needed separatly on variablebox and table.


likp:
@KentMossback is this change still valid?

KentMossback:
This has been implemented in another pr

PR #257, opened 2024-08-27
Patch/update dependencies

SjurSutterudSagen:
Since we want to keep the project up-to-date, we need to update the packages we use. This updates the packages used by Nx, and updates the other packages as allowed by package.json.

Tested to see if the problem with 'npm audit fix' was still there, but it is now fixed.

This PR should remove the current dependabot PRs we have.


MikaelNordberg (approved):
Looks good

PR #260, opened 2024-09-03
fix/px-web2-48: navigation rail checkup

SjurSutterudSagen:
Since we want the NavigationRail side menu to be done according to figma, we need to check that the work done for the first test version is in accordance with what is in figma. This is in terms of styling and functionality.

This fixes some styling, warnings in the dev tools in edge, the translations and tests for the NavRail.

It also adds the animations from Figma, which is why the framer-motion package is installed.


SjurSutterudSagen:
There is currently a bug when clicking on a selected item in the NavRail: The color changes to white before the hover takes effect. Have not been able to fix it, discussed it with the other devs, and we'll create a new task for that bug, since everything else in this check is done. And this code is needed for the NavBar task.

MikaelNordberg (approved):
Looks good

PR #261, opened 2024-09-09
Feature/test react router for loading

SjurSutterudSagen:
Since I want the others to look at this test of the react router for
loading the inital data, I need to clean up the branch a bit.

There are still problems to solve, but the basic loading of inital data
in parallell should now work. It also works with making selections for
new api calls, once the inital data has finished loading.

I did not spend much time trying to get Suspense to show, or styling
errors or similar. This is primarly for showing how  the routing/dataloading works.

Important note here: 
Up until now we have tried to not write code that is framework specific. While React Router is not a framework (yet, soon Remix), using it for loading data is very specific to it.


SjurSutterudSagen:
[first-load-large-table-react-router.json](https://github.com/user-attachments/files/16942554/first-load-large-table-react-router.json)
The performance of the new parallelized API calls. Load it into the performance tab of the browser dev tools.

SjurSutterudSagen:
Just noticed I also commented out the test for App in app.spec.jsx, so this also needs to be fixed.

SjurSutterudSagen:
I think this is too old and outdated compared to how the project is now. It's probably easier to just redo this work in the current code, than updating this PR. Closing this.

PR #265, opened 2024-09-10
Feature/pxweb2 278 restructure mainpage

MikaelNordberg:
The user interface now works better on differnt devices with small and big screens.


apps/pxweb2/src/app/app.module.scss
```diff
@@ -1,85 +1,105 @@
...
+  flex-direction: row;
 }
 
-.main {
+.mainContainer {
   display: flex;
-  width: 100%;
-}
+  background: var(--px-color-surface-subtle);
+  height: calc(100vh - 88px);
```
michaelpande, line 12: Test this on a smartphone. A couple of years ago when scrolling at a page the application header for the browser (where the URL to the website is) would appear in certain browsers. 

In other words, try scrolling up or down, check if application header appears.

**Why it matters:**
Earlier (maybe still) when the application header appears the view height of the page changes, as the application header takes up some of the view height.

michaelpande: It might not be a problem at all, just wanted to mention it

michaelpande (dismissed):
LGTM 👍 

Test the case I mentioned. 

I have not tested this PR myself, only read the code.

apps/pxweb2/src/app/app.module.scss
MikaelNordberg: When I scroll up and down on my iPhone I can see that the the header can be hidden/shown as you mention. It is very easy to display the header again by scrolling again. Do not know if this is an issue...

SjurSutterudSagen (dismissed):
LGTM

apps/pxweb2/src/app/app.module.scss
michaelpande: That's not the issue. The issue I was thinking about is that since the view height (vh) changes, items might jump around on the page. If that does not happen, then everything is good 👍 

I should have explained better. Was in another meeting so was a bit distracted, sorry 😄

SjurSutterudSagen (approved):
Good job, looks good to me.

PerIngeVaaje (approved):
OK

PR #266, opened 2024-09-11
Fix/pxweb2-140: NavigationBar checkup

SjurSutterudSagen:
Since we want to check the functionalty and styling of the Navigation
components that were made for the summer test version, we need to fix any issues we find.

This is a continuation of the work on the NavigationRail checkup, and should be merged after it. It also has the same bug with the animation.

Fixed styling and animation for the NavigationBar component.


MikaelNordberg (dismissed):
Looks good

MikaelNordberg (approved):
Looks good

PR #267, opened 2024-09-11
patch/Update dependencies

SjurSutterudSagen:
Since we want to keep the project up-to-date, we need to update our dependencies regularly. This updates the Nx dependencies, and the other dependencies within their allowed ranges.

While there is a currently task of replacing Nx with npm workspaces, I think we might still want to update this way until that task is done and merged.


PerIngeVaaje (approved):
OK

PR #294, opened 2024-10-08
Added a Dockerfile

likp:
Added a simple Dockerfile for PxWeb2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
-FROM node:20.11.1-slim AS build
+FROM node:20.18.0-slim AS build
 WORKDIR /app
 COPY . ./
 RUN npm install
 RUN npx nx reset
+RUN npm ci
 RUN npm run build
 
 # production environment
-FROM nginxinc/nginx-unprivileged
+FROM nginxinc/nginx-unprivileged::1.27-alpine3.20
 COPY --from=build /app/dist/apps/pxweb2 /usr/share/nginx/html
 EXPOSE 8080
 CMD ["nginx", "-g", "daemon off;"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM node:20.11.1-slim AS build
+WORKDIR /app
+COPY . ./
+RUN npm install
+RUN npx nx reset
+RUN npm run build
+
+# production environment
+FROM nginxinc/nginx-unprivileged
+COPY --from=build /app/dist/apps/pxweb2 /usr/share/nginx/html
+EXPOSE 8080
+CMD ["nginx", "-g", "daemon off;"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN npm ci
 RUN npm run build
 
 # production environment
-FROM nginxinc/nginx-unprivileged::1.27-alpine3.20
+FROM nginxinc/nginx-unprivileged:1.27-alpine3.20
 COPY --from=build /app/dist/apps/pxweb2 /usr/share/nginx/html
 EXPOSE 8080
 CMD ["nginx", "-g", "daemon off;"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
 FROM node:20.18.0-slim AS build
 WORKDIR /app
 COPY . ./
-RUN npm install
-RUN npx nx reset
 RUN npm ci
+RUN npx nx reset
 RUN npm run build
 
 # production environment
 FROM nginxinc/nginx-unprivileged:1.27-alpine3.20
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:20.11.1-slim AS build
+WORKDIR /app
+COPY . ./
+RUN npm install
+RUN npx nx reset
+RUN npm run build
+
+# production environment
+FROM nginxinc/nginx-unprivileged
```
runejo, line 9: I have no experiance with the unprivileged image but it looks promising. Mayby use a concret tag like `nginxinc/nginx-unprivileged:1.27-alpine3.20`

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:20.11.1-slim AS build
+WORKDIR /app
+COPY . ./
+RUN npm install
```
michaelpande, line 4: RUN npm ci

https://docs.npmjs.com/cli/v10/commands/npm-ci

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:20.11.1-slim AS build
```
michaelpande, line 1: Known vulnerabilities
image

michaelpande: https://hub.docker.com/layers/library/node/20.11.1-slim/images/sha256-2b3042f699adef6d192f2cdaf62eb1621d36792c2ab9f268b0a07745063bafad?context=explore

michaelpande (changes requested):
Should use an image without vulnerabilities, and npm ci 

Rest looks good 👍

Dockerfile
likp: Good catch copy-paste thing. I have upgraded to latest slime version 2.18.0

Dockerfile
likp: Have added this now

Dockerfile
likp: Ran with your suggestion.

michaelpande (dismissed):
LGTM 👍

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:20.18.0-slim AS build
+WORKDIR /app
+COPY . ./
+RUN npm install
+RUN npx nx reset
+RUN npm ci
+RUN npm run build
+
+# production environment
+FROM nginxinc/nginx-unprivileged::1.27-alpine3.20
```
SjurSutterudSagen, line 10: I think there is a ":" too many here. When i run the build command it fails at once, but when I remove one of the ":" it runs fine.

SjurSutterudSagen (changes requested):

Dockerfile
SjurSutterudSagen: I'm likely doing something wrong here, but I also can't seem to access the site when running the container locally.

likp: Good catch.I have pushed a change where I have removed the extra colon.

Could it be that you do not map the exported port to a local port on your machine?
I built the imagage with 

`docker build -t pxweb2:test .`

and ran it with 
`docker run --rm -p 8080:8080 pxweb2:test`

you could change the first `8080` port to something else if you already have something running on that port already.

SjurSutterudSagen: I ran 
` docker build . -t pxweb2/web`
and
`docker run -dp 8080:80 pxweb2/web`

I can't get access from the browser, but the command `docker exec -it musing_borg  curl localhost:8080` gave a response.

likp: I see you are mapping 8080 on your machine to container port 80. But the container is exposing the 8080 port (we are using the nginx unprivileged that uses port 8080 instead of 80 which require root access). 

That is also why curl works when you exec into the container. Because your hit the 8080 port where nginx is listening and serving the files.

SjurSutterudSagen: Thanks, that got it working. I was mixing up which side did what on the port mapping. Changed it and got it working. 

I did notice that refreshing the page, or manually changing the URL to something other than the base `localhost:8080/` gives me a 404 error.

runejo: I think the way to solve this is to use the `try_files` directive in nginx https://www.frontendundefined.com/posts/tutorials/nginx-react-router-404/

runejo: I have a working version, shall i commit it here?

likp: Yes, do it

likp: It solved the problem for me @SjurSutterudSagen could you also verify it?

runejo (approved):

Dockerfile
SjurSutterudSagen: Works fine here now

SjurSutterudSagen (approved):
LGTM

PR #297, opened 2024-10-14
Fix order of CSS in presentation component

SjurSutterudSagen:
The order of the CSS in the presentation component had some media queries before some other code. This counts as having nesting before non-nested CSS, and we get some spam in the console.

"Deprecation Warning: Sass's behavior for declarations that appear after nested rules will be changing to match the behavior specified by CSS in an upcoming version. To keep the existing behavior, move the declaration above the nested rule. To opt into the new behavior, wrap the declaration in `& {}`."

There is also some other spam in the console about SCSS, but it is not related to this. It seems to be related to some defaults being set in Vite.


PerIngeVaaje (approved):
OK

PR #304, opened 2024-10-15
Moved API URL setting to config.js

likp:
The API URL is set in the config.js file instead of being hard coded in OpenAPI.ts

### DIFF
--- a/libs/pxweb2-api-client/src/core/OpenAPI.ts
+++ b/libs/pxweb2-api-client/src/core/OpenAPI.ts
@@ -20,9 +20,9 @@ export type OpenAPIConfig = {
 };
 
 export const OpenAPI: OpenAPIConfig = {
-    BASE: 'https://api.scb.se/OV0104/v2beta/api/v2',
+    //BASE: 'https://api.scb.se/OV0104/v2beta/api/v2',
     //BASE: 'https://pxapi2-master-px.scb.se/api/v2',
-    //BASE: 'https://data.qa.ssb.no/pxapi2-beta/api/v2',    
+    BASE: 'https://data.qa.ssb.no/pxapi2-beta/api/v2',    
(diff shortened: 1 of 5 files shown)


libs/pxweb2-api-client/src/core/OpenAPI.ts
```diff
@@ -20,9 +20,9 @@ export type OpenAPIConfig = {
 };
 
 export const OpenAPI: OpenAPIConfig = {
-    BASE: 'https://api.scb.se/OV0104/v2beta/api/v2',
+    //BASE: 'https://api.scb.se/OV0104/v2beta/api/v2',
     //BASE: 'https://pxapi2-master-px.scb.se/api/v2',
-    //BASE: 'https://data.qa.ssb.no/pxapi2-beta/api/v2',    
+    BASE: 'https://data.qa.ssb.no/pxapi2-beta/api/v2',
```
SjurSutterudSagen, line 25: Should the URL that is uncommented here be the first one? The one to api.scb.se?

KentMossback (dismissed):
Looks good to me

libs/pxweb2-api-client/src/core/OpenAPI.ts
likp: Or hsould I remove all of them since the value is acctually pick set from the config.js file to prevent confusion?

SjurSutterudSagen: That might be better yes

likp: Unforotunatly you have to set BASE to something so I changed to to statistics swedens beta site and I also added a comment that it is overriden in config.js

SjurSutterudSagen (approved):
LGTM

PR #305, opened 2024-10-15
PXWEB2-118 First facelift

PerIngeVaaje:


libs/pxweb2-ui/src/lib/components/Table/Table.module.scss
```diff
@@ -63,3 +92,12 @@
...
 }
+tr.firstdim th{
+  border-top: 2px solid var(--px-color-border-default);
+  font-weight: bold;
+}
+tr.firstdim td{
+  border-top: 2px solid var(--px-color-border-default);
+  font-weight: bold;
+}
+
```
michaelpande, line 103: This should be inside .table { }

The reason is that .table make sure that it is scoped properly so that the styling does not affect other tr  og td elements. 

You use a class here (.firstdim), so it will probably never be an issue. 

When our SCSS is compiled the .table becomes .table-ei2849q or something like that. Which means that the changes that are scoped within that classname would only affect tables with that class. The reason it matters is that PxWeb2 might be placed in a context with other CSS and HTML elements not from the project itself. Styling td, tr, and other HTML elements directly is the same as styling them globally 😄 

This is a nitpick, so if it can't be solved by just moving it inside table, it can probably just stay as is.

PerIngeVaaje: Has put inside table

KentMossback (approved):
Looks good

PR #306, opened 2024-10-16
fix/px-web2-361: Fix languages sv and nor

SjurSutterudSagen:
The translations are a bit out of date, and not correct. This should fix this for all 4 translations files. I have only fixed existing translation keys in the application, and have not added any new translations from the confluence page.

I also expanded on the Modal and Select to be able to provide custom translations for the Modal buttons, since it used the `generic_buttons` translations, which do not have the same meaning in the Norwegian translation. This was only for the VariableBox.
```
EN = Save
NOR = Bekreft (confirm)
```

Important: The arabic translation file was very out of date, with many missing translations. This has been fixed with https://translate.i18next.com/, which uses google translate. It is not meant for any production use! We would need to do proper translations at a later point for that.

### DIFF
--- a/apps/pxweb2/public/locales/ar/translation.json
+++ b/apps/pxweb2/public/locales/ar/translation.json
@@ -1,55 +1,82 @@
 {
   "meta": {
-    "languageName": "العربية",
+    "languageName": "إنجليزي",
     "shorthand": "ar"
   },
   "common": {
     "title": "مرحبًا بك في PxWeb 2.0",
     "header": {
-      "title": "PxWeb 2.0",
-      "logo": "PxWeb 2.0"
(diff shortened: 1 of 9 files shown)


SjurSutterudSagen:
Can't log in to cloudflare right now. Bitwarden is down and I need to log in. Will do it tomorrow.

likp:
@SjurSutterudSagen I get that same build error when trying to build PxWeb2 as Cloudflare. Can you give it a look.

SjurSutterudSagen:
I forgot to update the i18n-interface types, which made the build fail on a type error. Fixed and builds fine now.

likp (approved):
Looks good

PR #307, opened 2024-10-17
Patch/update project dependencies

SjurSutterudSagen:
To keep the project up-to-date we need to update our dependencies. This
updates the Nx dependencies and the project dependencies within the
ranges in package.json. It does not update the dependencies outside of
those.

The newest version of Nx is 20.0.1, but there are a lot of issues and bugs with that newest major version. Opted therefore to only update to the newest version of Nx 19.

"Npm update" added a license field to all the dependencies. Seemed to be a change they made in the NPM CLI.

Co-authored-by: Øyvind Edvardsen <>


oyessb (approved):
👍

PR #316, opened 2024-10-22
Patch/update project dependencies

SjurSutterudSagen:
This updates the Nx dependencies and the dependencies in the package.json within the allowed ranges set there.

I have still not updated to the newest Nx major version, since it seems there are still serious bugs with it. Looked at the "Issues" and "PRs" on their repository.


PerIngeVaaje (approved):
OK!

PR #318, opened 2024-10-28
Sticky items in Virtuoso variabelbox

oyessb:
Added value for topItemCount prop in Virtuoso to make search and select all sticky if shown.


SjurSutterudSagen:
Should this PR also have the animations for the sticky items?

PR #327, opened 2024-10-31
Feature/Add showing the codelist from the defaultSelection in the VariableBoxes

SjurSutterudSagen:
Since the API now supports a default codelist in the defaultSelection of a table, we need the implementation for showing this when loading a table.

This work was started as part of PxWeb2-198, but was moved to it's own task and branch. Any remaining cleanup will be done during that task.


PerIngeVaaje (approved):
Good

PR #328, opened 2024-11-01
Update and migrate Nx to latest major version

SjurSutterudSagen:
This updates Nx to the latest version, and runs the necessary migrations. I tried updating to the latest version of 19, but our test command stopped running then. Checking the logs it seemed to be a problem with some db connections in the nx system. Therefore I tried updating to the latest major version, even though it has seemed to have some bugs, going by their github repository.

Tested running tests, build, start etc. Also looked at the app itself, and I do not find any issues.


PerIngeVaaje (approved):
OK

PR #329, opened 2024-11-04
PXWEB2-118 If there is no stub left columns in header and body gets d…

PerIngeVaaje:
…ouble columns.This is wrong. Should now be corrected


SjurSutterudSagen (approved):
Looks good! Nice catch!

PR #333, opened 2024-11-06
Sticky items for search/select are added by topItemCount prop. When s…

oyessb:
…crolling down a custom TopItemList i used to override sticky div with empty fragment.


SjurSutterudSagen (approved):
Looks good, behaves as expected when scrolling, except that one bug when scrolling very slowly. But as I said, I think that one can be fixed when looking into animations. Might be quite technical with virtuoso, and no point spending loads of time now fixing it, if we can't get the animations to work properly anyways, or the other way around.

PR #334, opened 2024-11-11
Pxweb2 384 fix unit tests for variable box

michaelpande:
preload fonts and fix unit tests for variable box so that it uses a mocked version of react virtuoso


SjurSutterudSagen (approved):
Looks good to me!

PR #336, opened 2024-11-12
Feature/pxweb2 198 add ability to change codelist on variable

SjurSutterudSagen:
This adds an API call that returns the new values for a chosen codelist
in the VariableBox, and plugs that into the existing code for showing
new codelists. This means the VariableBox and state now can handle
changing codelists and rendering them.

I also formatted some files.


SjurSutterudSagen:
It's best to wait with this one so we can get the API fix merged into it. Will make it much easier to test, when the values are actually correct from the beginning.

PerIngeVaaje (approved):
OK!

PR #337, opened 2024-11-13
Feature/pxweb2 319 add metainfo to cells

PerIngeVaaje:
Added metainfo used for screenreaders


oyessb (approved):
Tested with NVDA. Reads table as expected 👍  
Note about empty table header is added in Jira.

PR #338, opened 2024-11-13
PXWEB2-389 Dynamic heights for variable boxes

oyessb:
Set height for Virtuoso list to calculated total height if total is less than 380px high.


PerIngeVaaje (approved):
Looks OK!

PR #342, opened 2024-11-15
Fixed most issues

michaelpande:
Makes the items in the list look like this:
image

also changed the height of the element to select groups.


MikaelNordberg (dismissed):
Looks good

PR #343, opened 2024-11-18
Changed focus order and styling for variable box

oyessb:
Changed focus order for variable box.
Added focus styling for chevron. 
Changed style for select so it is not hidden for focus-visible state.


michaelpande (approved):
LGTM 👍 

Har ikke testet det, kun lest koden

PR #345, opened 2024-11-21
Pivoted table on mobile

MikaelNordberg:
First version of mobile table


SjurSutterudSagen (approved):
Looks good. Can improve some parts later, like pivotCW name and logic when we add that later.

PR #356, opened 2024-11-26
Update the API-client to the newest version

SjurSutterudSagen:
Updates the API client to the newest version, since we want to update the API.


likp (approved):
LGTM

PR #374, opened 2024-11-27
Remove nx

michaelpande:
Removed Nx from most places, there are still some places left. Where we use nx as a plugin or nx/react or stuff like that. Making this PR to get errors in the build steps in order to get closer to something that can be merged. 

After it is merged the rest can easier be handled. 

The worry is that the longer this task isn't merged the worse the merge will be. It's guaranteed to be pretty horrific already 🙈


PerIngeVaaje (dismissed):
OK

MikaelNordberg (approved):
Fine

PR #379, opened 2024-12-02
Build multi-platform docker image on every PR

runejo:


likp (approved):
LGTM , but the CI build is failing

Dockerfile
```diff
@@ -1,12 +1,11 @@
...
 WORKDIR /app
 COPY . ./
-RUN npm ci \
-  && npm run build
+RUN npm ci && npm run build
 
 # production environment
 FROM nginxinc/nginx-unprivileged:1.27.2-alpine3.20
-COPY --from=build /packages/pxweb2/dist /usr/share/nginx/html
+COPY --from=build /app/packages/pxweb2/dist /usr/share/nginx/html
```
michaelpande, line 8: This should be wrong, is the build step building to the wrong location perhaps? It should not be necessary to add /app/ here 🤔

runejo: I was failing on COPY, but it could also be fixed like this, without leading slash
`COPY --from=build packages/pxweb2/dist /usr/share/nginx/html`

likp: It should be correct the code is placed in /app.
When is test localy the build does not fail.

PR #380, opened 2024-12-02
Fix building and testing on macs

SjurSutterudSagen:
Removing the Nx packages broke the build step on arm processors, due to the bug in npm with optional architecture dependencies. The same bug I've ran into several times when updating the Nx packages. This fixes those missing dependencies (had to delete node_modules and package-lock.

Also, running the tests after removing Nx, in a clean repo, failed on TypeErrors with getTextBy. This adds a temporary fix by downgrading testing-library/react to the previous major version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
   "private": true,
   "dependencies": {
     "@uidotdev/usehooks": "^2.4.1",
-    "react-virtuoso": "^4.12.1"
+    "react-virtuoso": "^4.12.1",
+    "sass": "^1.81.0"
   },
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "7.17.0",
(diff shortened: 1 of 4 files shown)


michaelpande (approved):

runejo:
Nice, this also fixed the new workflow in  #379 👍

package.json
```diff
@@ -15,7 +15,8 @@
   "private": true,
   "dependencies": {
     "@uidotdev/usehooks": "^2.4.1",
-    "react-virtuoso": "^4.12.1"
+    "react-virtuoso": "^4.12.1",
+    "sass": "^1.81.0"
```
runejo, line 19: Should be under devDependencies.
Yesterday morning we talked about using https://www.npmjs.com/package/sass-embedded is it a better option? @michaelpande @SjurSutterudSagen

PR #384, opened 2024-12-03
Dependabot version updates - Support NPM Workspaces and Docker

runejo:
The `lockfile-only` was for NX compatibility i think


michaelpande (approved):
LGTM 👍

PR #388, opened 2024-12-04
Use motion instead of framer-motion

michaelpande:


SjurSutterudSagen (approved):
LGTM, good that we are making the switch to the new open source version!

PR #405, opened 2024-12-04
Use react-router instead of react-router-dom

michaelpande:


SjurSutterudSagen (approved):
Looks good to me, tested in the browser with what the pipeline built.

PR #433, opened 2024-12-13
Fix/pxweb2 437 dataprovider bug

MikaelNordberg:
Changed how the DataProvider determines if data cells asked for by the user are already loaded in the accumulated data cube or not. Now the DataProvider checks cell by cell in the accumulated data cube.


MikaelNordberg:
Changed how the DataProvider determines if data cells asked for by the user are already loaded in the accumulated data cube or not. Now the DataProvider checks cell by cell in the accumulated data cube.

likp (approved):
LGTM

PR #434, opened 2024-12-16
Remove Nx Console from recommended repo extensions

SjurSutterudSagen:
Since we have removed Nx from the repository, we should not recommend the Nx Console extension for VSCode.


MikaelNordberg (approved):
Looks good

PR #438, opened 2024-12-17
Update navigtionDrawer to start closed on mobile

SjurSutterudSagen:
This changes the app so when initially loading the site on mobiles or tablets, the navigationDrawer starts closed.

It also adds a top-level AppProvider and context to keep the states for isMobile and isTablet. The logic for calculating this has been moved from app.tsx, and expanded. We have talked about needing this earlier, and seemed to be a good time to add it now.

Also found that the current checks for the context being inside their providers do not work. So added a working solution for this in the new provider. I'm not sure if this is the best way of doing this though.


michaelpande (dismissed):
LGTM 👍

PerIngeVaaje (approved):
ok

PR #460, opened 2025-01-08
Fix/pxweb2-198: Show groupings last in codelists

SjurSutterudSagen:
This moves the groupings last when there are both groupings and value
aggs among the codelists. Also added some tests, and minor refactor.


michaelpande (approved):
LGTM 👍 🚀

MikaelNordberg (approved):
Looks good

PR #482, opened 2025-01-20
Validate HTML after render

michaelpande:
image


runejo (approved):
Works on my mac after running `npx playwright install` 👍

SjurSutterudSagen:
> Works on my mac after running `npx playwright install` 👍

Same here, though I did notice that sometimes i get 76 errors, like in the picture, and sometimes 663. Maybe it sometimes runs the checks without the table?

PR #486, opened 2025-01-27
fix/Add storybook-static to lint ignore

SjurSutterudSagen:
This fixes an issue where sometimes the output directory for storybook isn't moved. Which meant the lint rules that run before commits would lint inside the generated storybook folder. This was a problem since there were many issues to our lint rules in there. It was allready ignored by git, so I see no reason to not just ignore it for linting too.


MikaelNordberg (approved):
Looks good

PR #489, opened 2025-01-28
Update the generated api-client to the latest spec

SjurSutterudSagen:
This updates the generated api-client files to the latest spec.


likp (approved):
Looks great!

PR #491, opened 2025-01-28
Fix missing type-module in package.json files

SjurSutterudSagen:
This adds the line '"type": "module"' to the package.json files that are missing it as of now. It cleans up the messages we get when we build the projects. There were several places where it complained that we built it using the old "require()" way. This made reading the terminal output more difficult than it needs to be.

It also adds some formatting to one vite.config.ts file, but nothing functional in it has changed.

Not sure if we need the '"type": "module"' line in the root project, so leaving it for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
   "name": "@pxweb2/source",
   "version": "0.0.0",
   "license": "MIT",
-  "type": "module",
   "scripts": {
     "prepare": "husky",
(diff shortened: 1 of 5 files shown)


michaelpande (approved):
LGTM 👍

michaelpande:
It's not necessary to add type module to the parent package.json

michaelpande (approved):

PR #492, opened 2025-01-28
Remove the last types from NX in the projecct

SjurSutterudSagen:
This fixes the errors we have in the tsconfig files from the old types from NX that were left. After this there should not be many more times where we have the red 'error' indicator in our IDEs from config files. One exception is the one file with a test part, that causes an error.

### DIFF
--- a/packages/pxweb2-ui/tsconfig.lib.json
+++ b/packages/pxweb2-ui/tsconfig.lib.json
@@ -4,8 +4,6 @@
     "outDir": "./dist/out-tsc",
     "types": [
       "node",
-      "@nx/react/typings/cssmodule.d.ts",
-      "@nx/react/typings/image.d.ts",
       "vite/client"
     ]
   },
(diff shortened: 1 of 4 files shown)


packages/pxweb2-ui/tsconfig.lib.json
```diff
@@ -4,8 +4,6 @@
     "outDir": "./dist/out-tsc",
     "types": [
       "node",
-      "@nx/react/typings/cssmodule.d.ts",
```
michaelpande, line 7: Do we not get errors when this is removed? If that is the case. Great! Remove it 👍

michaelpande (approved):
Just one comment, if that is not necessary anymore, feel free to remove it 🚀

packages/pxweb2-ui/tsconfig.lib.json
SjurSutterudSagen: Yea, I didn't see any errors. And i could build, test and run it fine so :)

PR #493, opened 2025-01-28
Add mocking in test for useTranslate

SjurSutterudSagen:
This fixes the spam we have had in the terminal related to react-i18next and the useTranslate function. It has been quite spammy, making the testing results hard to read.

Took some troubleshooting to get it to work, but made a mocking function in the ui library, and the export it to the web app to use it there too. The lines for importing and using them are at the top of the files, because they need to be ran before importing the other components that use the function. This seems clearer and easier to deal with than trying to figure out which import line needs to be moved below the others.

Also fixed some small issues with missing variables in the tests where I saw them.

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,16 @@
+# Path to sources
+#sonar.sources=.
+sonar.exclusions="**/**.spec.**"
+#sonar.inclusions=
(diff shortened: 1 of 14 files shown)


michaelpande (approved):
LGTM 👍

PR #502, opened 2025-01-30
Fix vite/vitest configs and remove useTranslate mock

SjurSutterudSagen:
This fixes the configs for vite and vitest, which has errors in them due to a incorrectly places "test" part. Now we merge them correctly according to the docs.

This also removes the mocking of the useTranslate function, since that makes the dev environments crash. I do think we might be able to mock them now, if we use a setupTests file, and import it in both the vitest config files. However, since the main branch now is broken for dev, I prioritize only fixing that in this branch.


michaelpande (approved):
🚀

PR #503, opened 2025-01-30
Fix mocking of react-i18next's useTranslate

SjurSutterudSagen:
This is the second attempt at mocking the useTranslate function from react-i18next. The first time broke the dev environment, but lead to other fixes and to the beginning of this fix. Adds a setupTests file to both the web application and the ui-library. Tested it, and it seems to work fine now. Dev environment runs fine, and the terminal is not spammed with react-i18next errors when running tests.

Please do test this, since we (I) do not want to break something in main again :D


MikaelNordberg (approved):
Seems to work fine

PR #505, opened 2025-01-30
Finishing touches 2

michaelpande:
Fixes focus bug in meny & makes sure the provider is used correctly


SjurSutterudSagen (approved):
Looks good, tested both the focus logic and the provider test.

PR #508, opened 2025-01-31
Add storybook addon for a11y checks in components

SjurSutterudSagen:
This adds the storybook addon for a11y checks in the component views in storybook itself. This was installed before, but seems to have gone missing at some point.


michaelpande (dismissed):

MikaelNordberg (approved):
Looks good

PR #509, opened 2025-01-31
Last focus fix

michaelpande:
Focus should now look as intended in variablebox
Focus should now work as intended in variablebox
Fixed pretty unfortunate bug in which colliding Ids made the focus jump from one variablebox to another


oyessb (approved):
👍

PR #518, opened 2025-02-03
Update Vitest from v2 to v3

SjurSutterudSagen:
This updates Vitest to the latest major version, v3. There were no code changes needed for us, according to the migration guide, and everything runs as expected.


SjurSutterudSagen:
Closing this, i forgot we can use the dependabot PR instead.

PR #519, opened 2025-02-04
Fix/cleanup tests output to increase readability

SjurSutterudSagen:
This fixes the spam we have when we run tests in the terminal. It fixes the issues in the test files, that caused said spam. And it adds a check for if the code is run during tests to the console.logging in the accessibility provider.


MikaelNordberg (approved):
Looks good

PR #522, opened 2025-02-05
Fix PxWeb2 font-family and font-weights

SjurSutterudSagen:
This fixes how we create our own font-family in the project. The font-family was initially created so we had 3, one for each font-weight. This leads to issues with browser user-agents, that automatically try to add boldness etc. to h1 tags and similar. For us, all the titles were too bold. Now the font-families are merged into one, with the font-weights: 400, 500 and 700. The site looks the same in chrome and safari.

### DIFF
--- a/packages/pxweb2-ui/.storybook/main.js
+++ b/packages/pxweb2-ui/.storybook/main.js
@@ -22,5 +22,8 @@ const config = {
     name: getAbsolutePath('@storybook/react-vite'),
     options: {},
   },
+  staticDirs: [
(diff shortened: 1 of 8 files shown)


SjurSutterudSagen:
So, they weren't false positives as such. Storybook couldn't load our static font files, so we needed to set up loading those fonts in the Storybook config. Now it loads everything correctly, and Chromatic found no differences. Ready for review.

oyessb (approved):
Looks good 👍

PR #523, opened 2025-02-05
PXWEB2-413 Screen reader info for variable box

oyessb:
Added screen reader information when focusing on accordion for variable box


PerIngeVaaje (approved):
OK

PR #525, opened 2025-02-05
PXWEB2-537 Change value for checkbox with space

oyessb:
Refactor Checkbox component to handle space key for toggling only


MikaelNordberg (approved):
Looks good

PR #527, opened 2025-02-05
Feature/pxweb2 428 jsonstat2 mapping

MikaelNordberg:
Part 1 of the jsonstat2 mapping.
- Will replace the existing metadata mapping using TableMetadataResponse
- Functionality developed this far for PxWeb 2 will work.


packages/pxweb2/src/mappers/JsonStat2ResponseMapper.ts
```diff
@@ -145,17 +158,71 @@ function mapDimension(id: string, dimension: any, role: any): Variable | null {
...
   return variable;
 }
 
+/**
+ * Maps the extension object of a dimension from a JSON-stat 2.0 response to a Variable object.
+ *
+ * @param extension - The extension object from the JSON-stat 2.0 response.
+ * @param variable - The Variable object to which the extension should be added.
+ */
+function mapDimensionExtension(extension: any, variable: Variable): void {
```
SjurSutterudSagen, line 183: I would really like it if we don't have 'any' if we can avoid it, is there a type you can narrow this down to later? So, set it to unknown first, then you would need to type it before you can use it later.
```suggestion
function mapDimensionExtension(extension: unknown, variable: Variable): void {
```

SjurSutterudSagen (changes requested):

SjurSutterudSagen (dismissed):
Looks good. As we discussed, the 'any' types we will fix later.

runejo (dismissed):

SjurSutterudSagen:
Setting it back to draft since Micke is sick, but the testers found something wrong. To avoid accidental merging this.

SjurSutterudSagen (approved):
LGTM

PR #528, opened 2025-02-07
Update ariaControls to include uniqueId

oyessb:


SjurSutterudSagen (approved):
Looks good! Fixes the error according to axe devtools

PR #529, opened 2025-02-07
Prevent autoscrolling when opening variableboxes

SjurSutterudSagen:
Since we have implemented custom keyboard navigation for opening/closing the variableboxes, we have to prevent that function from triggering the normal auto-scrolling for the spacebar.


oyessb (approved):
Looks good 👍

PR #530, opened 2025-02-07
Remove aria-label from header div without role

oyessb:


SjurSutterudSagen:
Do we need the translations that are left here? Or should those also be removed?

PerIngeVaaje (approved):
OK

PR #535, opened 2025-02-11
Add keyboard nav and focus to the Select in VarBox

SjurSutterudSagen:
This adds the keyboard navigation and focus marking to the Select and modal in the variableBox. It refocuses the Select when the modal closes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ theme/variables.css
 
 vite.config.*.timestamp*
 *storybook.log
+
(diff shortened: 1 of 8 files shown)


SjurSutterudSagen:
What is left is to look into a solution for the Enter key when closing the modal, since that does not refocus correctly. Maybe there is something special about the Enter key and how browsers handle it.

KentMossback (approved):
Looks good!

PR #547, opened 2025-02-20
Fix missing contacts in table testdata

SjurSutterudSagen:
This fixes an error when building due to a missing field in the testdata for the Table component.


MikaelNordberg (approved):
Looks good

PR #548, opened 2025-02-20
Feature/pxweb2 428 jsonstat2 notes

MikaelNordberg:


packages/pxweb2/src/mappers/JsonStat2ResponseMapper.ts
```diff
@@ -189,47 +261,148 @@ function mapDimension(id: string, dimension: any, role: any): Variable | null {
...
+  dimensionExtension: extension_dimension,
+  code: string,
+  noteIndex: number,
+): boolean {
+  if (dimensionExtension?.categoryNoteMandatory) {
+    if (dimensionExtension.categoryNoteMandatory[code]) {
+      if (dimensionExtension.categoryNoteMandatory[code][noteIndex]) {
+        return true;
+      }
+    }
```
SjurSutterudSagen, line 369: Can this be simplified with optionals in the last if? Instead of 3 ifs?

packages/pxweb2/src/mappers/JsonStat2ResponseMapper.ts
```diff
@@ -189,47 +261,148 @@ function mapDimension(id: string, dimension: any, role: any): Variable | null {
...
+  if (dimensionExtension?.show) {
+    if (dimensionExtension.show === 'code') {
       return 'code';
-    } else if (dimension.extension.show === 'value') {
+    } else if (dimensionExtension.show === 'value') {
       return 'value';
-    } else if (dimension.extension.show === 'code_value') {
+    } else if (dimensionExtension.show === 'code_value') {
       return 'code_value';
     }
```
SjurSutterudSagen, line 407: More a preference comment, the "else if"s here are not really needed, since there are returns inside of each. Personally I prefer pure ifs, but this is subjective I think.

SjurSutterudSagen (changes requested):
Good job, had some comments.

SjurSutterudSagen (commented):
LGTM :)

SjurSutterudSagen (approved):

PR #550, opened 2025-02-24
Remove alpine version in nginx-unprivileged

runejo:
Dependabot will not upgrade `1.27.3-alpine3.20-slim` to `1.27.4-alpine3.21-slim` so we remove the alpine version number


SjurSutterudSagen (approved):
LGTM

PR #556, opened 2025-02-26
Fix wrong fill and clip rule syntax in new icons

SjurSutterudSagen:
This fixes the fill-rule and clip-rule warnings in the test command. It should fix them for all our currently implemented code, since I ran search and replace on the entire codebase (except some Storybook specific things).


PerIngeVaaje (approved):
OK

PR #559, opened 2025-02-27
Feature/pxweb2 438 sidesheet notes

MikaelNordberg:


packages/pxweb2/src/app/components/TableInformation/Notes/NotesTab.tsx
```diff
@@ -0,0 +1,176 @@
+import cl from 'clsx';
+import useVariables from '../../../context/useVariables';
```
MikaelNordberg, line 2: Fix

packages/pxweb2-ui/src/lib/components/Notes/noteCollection.ts
```diff
@@ -0,0 +1,255 @@
+import { Note } from '../../shared-types/note';
+import { PxTableMetadata } from '../../shared-types/pxTableMetadata';
+import { Value } from '../../shared-types/value';
+import { Variable } from '../../shared-types/variable';
+import { VartypeEnum } from '../../shared-types/vartypeEnum';
+
+// Internal type holding mandatory and non mandatory notes for a table
+export type tableNoteCollection = {
+  //SymbolExplanationNotes: string[];
```
MikaelNordberg, line 9: Fix

SjurSutterudSagen (approved):
Good job! Still needs to expand on the tests and add story so we can see it in Storybook/Chromatic. If you could create new tasks for those, then this one can be merged. 

Approving so we can get it merged before the holidays (just create those 2 tasks first).

michaelpande (approved):
No big things, but some comments about readability and complexity. Fix what you agree with

packages/pxweb2-ui/src/lib/components/Notes/VariableNotes.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }
+  const returnString = structuredClone(string);
+  return returnString.charAt(0).toUpperCase() + returnString.slice(1);
+}
+
+export function VariableNotes({
+  variableNotes,
+  showVariableName = true,
+}: VariableNotesProps) {
+  return (
```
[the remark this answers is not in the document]
MikaelNordberg: Good point. I make a note to fix this as a separate PR later on. Need to merge this code now.

PR #565, opened 2025-03-04
PXWEB2-452 Override apiUrl with query-string parameter

likp:
This PR makes it possible to override the apiUrl configuration setting with a query string parameter called `apiUrl`

e.g. one could write ...


MikaelNordberg (approved):
Looks good

SjurSutterudSagen:
Is this supposed to be a permanent solution, so anyone can do this on a PxWeb2 installation? If so we might need to add validation of the API response for added security.

likp:
> Is this supposed to be a permanent solution, so anyone can do this on a PxWeb2 installation? If so we might need to add validation of the API response for added security.

No, this should not be a permanent solution. It is more of a temporary quick fix during development.

PR #567, opened 2025-03-05
Feature/pxweb2-460: Add global color values and fix semantic colors in the default theme

SjurSutterudSagen:
This adds some new global and semantic colors. It also fixes all the missing, old global colors, and makes all the semantic colors reference the global colors instead of having their own color codes.

Refactor the theme file to add sections like "background" instead of each semantic color being
on its own line with a "background-" prefix. It now also matches how they are in Figma from the designer.

Also removed some unused semantic colors.

Lastly, SonarQube complained about the key in the Highlight component, so asked "Claude 3.7 Sonnet" to fix it.


michaelpande (approved):
LGTM 👍

PR #568, opened 2025-03-06
Patch/update minor and patch dependencies

SjurSutterudSagen:
This updates the projects packages with currently available minor and patch dependency updates.

When it comes to the Storybook v8.5.8 -> v8.6.4 update:
Storybook recommends using 'react-docgen' since it is faster, but it is not accurate enough for our use-case. Union types were shown as [union], instead of the different values. Changed it back to 'react-docgen-typescript'.


MikaelNordberg (approved):
Looks good

PR #578, opened 2025-03-11
Feature/px web2 540 fix keyboard navigation navbar navrail

SjurSutterudSagen:
This updates the NavigationBar/Rail to handle keyboard navigation without needing any tabIndex changes. The components and app.tsx were refactored so the HTML barely changes when switching between the horizontal and vertical versions. Also added a 
heck for if the SkipToMain is focused, so it can be used in calculating heights for elements.

Also, we might consider refactoring the NavigationRail/Bar into one component, since the functionality is almost 100% the same. But have not done that as of now, since it might be a bigger jobb in fixing the CSS for that.

There are also various smaller fixes for removing old commented out code/fixing any types etc., that were done in the files I was editing anyways.

(This is only part of the PxWeb2-540 task, since I'm trying to keep the PRs smaller.)


michaelpande (approved):
LGTM 👍

michaelpande:
I think joining the components potentially could work, however, if there is a lot of little differences we have to account for by adding IF/ELSE everywhere, then having two is more flexible

SjurSutterudSagen:
The differences between the NavRail and NavBar I think are only to stop SonarQube from complaining about duplicate code. But would need to just recheck it to be sure.

PerIngeVaaje (approved):
OK

PR #588, opened 2025-03-18
PxWeb2-540: Fix firefox focusing a container div with kb-nav

SjurSutterudSagen:
This fixes the focus order in firefox, when using a keyboard to navigate. Tested and it does not affect Safari or Chrome.


oyessb (approved):
👍

PR #591, opened 2025-03-18
Fix/install specific versions of react typescript

SjurSutterudSagen:
This installs specific versions of React, the React types and TypeScript in the root of the project. We can move them into projects later, if a project needs a different version.


MikaelNordberg (approved):
Looks good

PR #593, opened 2025-03-19
Fix/pxweb2-540: Two small fixes for keyboard nav/focus marking

SjurSutterudSagen:
This fixes:

- Position on the NavigationDrawer when skipToMain is focused on mobile
- Autoscrolling when spacebar is used to click the Hide button in the VariableBox


michaelpande (approved):
LGTM 👍

PR #605, opened 2025-03-27
Remove old comments that override Intl options ts

SjurSutterudSagen:
This removes old comments that override the typechecking for Intl options in the i18next formatters. They are no longer needed since the correct types seem to be updated in TypeScript now (they were missing before).


MikaelNordberg (approved):
Looks good

PR #606, opened 2025-03-27
Fix the spacing in TableView for main content

SjurSutterudSagen:
This re-adds the gap for the contentAndFooterContainer for desktops, to fix the current bug where this gap is missing.


PerIngeVaaje (commented):
OK

PR #608, opened 2025-03-27
Add maxDataCells to web app config

SjurSutterudSagen:
Since we need to have a number that indicates the maximum number of cells we support in the TableView component, and the api-calls it makes. We need to add a field to the web app config that can be configured and used for this purpose.

Later we might add functionality so that this setting can be directly gotten from the api also/instead.

This is only adding this field, not using it anywhere yet. But I tested it and made sure the config validation fails if it is missing.


PerIngeVaaje (approved):
OK

oyessb (approved):
👍

PR #609, opened 2025-03-31
Added new BottomSheet component. Implemented in TableInformation component for mobileview.

oyessb:


packages/pxweb2-ui/src/lib/components/BottomSheet/BottomSheet.module.scss
```diff
@@ -0,0 +1,112 @@
...
+  align-items: center;
+  flex-shrink: 0;
+  margin-bottom: 0;
+
+  background: var(--px-color-surface-default);
+  border-radius: var(--px-border-radius-large) var(--px-border-radius-large) 0px
+    0px;
+
+  // Not from Figma
+  padding: 0px 0px 0px 0px;
```
michaelpande, line 17: Padding: 0;

packages/pxweb2-ui/src/lib/components/BottomSheet/BottomSheet.module.scss
```diff
@@ -0,0 +1,112 @@
+@use '../../../../style-dictionary/dist/scss/fixed-variables.scss' as fixed;
+
+.bottomSheet {
+  display: flex;
+  width: 100%;
+  height: 95vh;
```
michaelpande, line 6: This might need to be tested on mobile browsers. As the viewport changes when scrolling on certain browsers. It used to be like that at least, maybe it's changed now. 

Many mobile browsers used to change the viewport height when the user scrolls. Because the interface of the browser gets shown when scrolling, and gets hidden incertain cases. That used to change the viewport height. That makes the site move the elements around when the user stops scrolling.

packages/pxweb2-ui/src/lib/components/BottomSheet/BottomSheet.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const bottomSheetRef = useRef<HTMLDialogElement | null>(null);
+  const [isClosing, setIsClosing] = useState(false);
+
+  useEffect(() => {
+    setIsBottomSheetOpen(isOpen);
+  }, [isOpen]);
+
+  useEffect(() => {
+    const modalElement = bottomSheetRef.current;
+    if (modalElement) {
```
michaelpande, line 36: Consider negating this and returning, instead of nesting the code within


If(!modalElement) {
  return
}

michaelpande (approved):
Fix the things you agree with, and try to test on both iPhone and a Chromium based Android browser.

PerIngeVaaje (approved):
OK

packages/pxweb2-ui/src/lib/components/BottomSheet/BottomSheet.module.scss
oyessb: Fixed the other comments. We have tested on Iphone and samsung. The height seems to be stable on both.

PR #610, opened 2025-03-31
chore: upgrade minor/patch versions of dependencies

Glenruben:
Used NCU to perform all available package upgrades Tested and ran locally


SjurSutterudSagen (approved):
LGTM

PR #613, opened 2025-04-01
SonarQube CI-based analysis with coverage report

runejo:
Switch to SonarQube CI-based analysis so that we can provide a test coverage report


runejo:
Works on my fork https://sonarcloud.io/summary/overall?id=runejo_PxWeb2&branch=main but need admin role on sonarqube to make necessary changes

SjurSutterudSagen:
I'm for this, but the api-client is generated from the spec, so should that be included here? Or should it be ignored?

runejo:
> I'm for this, but the api-client is generated from the spec, so should that be included here? Or should it be ignored?

I agree, it should be ignored, I've updated the config now 👍

.github/workflows/ci-tests.yml
```diff
@@ -22,11 +22,20 @@ jobs:
...
+        with:
+          fetch-depth: 0
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
       - run: npm ci
       - run: npm run build
-      - run: npm test
```
michaelpande, line 32: Will tests still be run? Before this change we run "npm test" on all the packages.

If yes, then it's all good 🚀

runejo: Yes `npm test` wil run just as before and `npm run coverage` will run the tests **and** make the coverage reports. 

(I temporary changed one of the tests just to make sure ;-) )

SjurSutterudSagen (approved):
LGTM, will be nice to get the coverage!

PR #614, opened 2025-04-01
Fix for checking if accumulated data is valid

MikaelNordberg:
Fix for checking if accumulated data is valid or not


likp (approved):
LGTM

PR #615, opened 2025-04-01
fix: Variable's values' txts when changing codelists

SjurSutterudSagen:
Since we want the ability to change codelists, we need to update the function that parses the new codelist metadata for a variable to properly handle the ValueDisplayType and value label texts. This adds similar parsing as the Json2Mapper.

Small refactor to extract shared functionality. If we decide to redo/change how we get the metadata when changing codelists, some of this might have to be redone/undone again.

### DIFF
--- a/packages/pxweb2/src/app/util/utils.ts
+++ b/packages/pxweb2/src/app/util/utils.ts
@@ -1,0 +1,23 @@
+import { ValueDisplayType } from '@pxweb2/pxweb2-ui';
+
+/**
+ * Returns the label text for a value based on the value display type.
+ *
+ * @param valueDisplayType - The value display type for the dimension.
+ * @param code - The code of the value.
+ * @param label - The label of the value.
+ * @returns The label text for the value.
+ */
+export function getLabelText(
+  valueDisplayType: ValueDisplayType,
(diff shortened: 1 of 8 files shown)


packages/pxweb2/src/app/util/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @param code - The code of the value.
+ * @param label - The label of the value.
+ * @returns The label text for the value.
+ */
+export function getLabelText(
+  valueDisplayType: ValueDisplayType,
+  code: string,
+  label: string,
+): string {
+  if (valueDisplayType === 'code') {
```
michaelpande, line 16: Due to multiple return statements, there is no need to use else. Very nitpicky and mostly a matter of preference. So up to you 😊

It can be simplified to:

If(){Return} 
If(){Return} 
Return

michaelpande (approved):

packages/pxweb2/src/app/util/utils.ts
SjurSutterudSagen: I agree, and I also prefer that so I'll update this.

PR #616, opened 2025-04-03
Add subtle variant to Tag component and adjust styles for size variations

Glenruben:
Should be ready for review

### DIFF
--- a/packages/pxweb2-ui/src/lib/components/Tag/Tag.module.scss
+++ b/packages/pxweb2-ui/src/lib/components/Tag/Tag.module.scss
@@ -8,18 +8,33 @@
 }
 
 .medium {
-  padding: 4px 8px;
+  height: 32px;
+  padding: 4.5px 8px 4px 8px;
   border-radius: var(--px-border-radius-small);
 }
 
 .small {
-  padding: 1px 6px 2px 6px;
+  height: 24px;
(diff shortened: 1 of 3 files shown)


packages/pxweb2-ui/src/lib/components/Tag/Tag.module.scss
```diff
@@ -8,20 +8,35 @@
...
-  padding: 4px 8px;
+  height: 32px;
+  padding: 4.5px 8px 4px 8px;
   border-radius: var(--px-border-radius-small);
 }
 
 .small {
-  padding: 1px 6px 2px 6px;
+  height: 24px;
+  padding: 1.5px 4px 3px 4px;
```
michaelpande, line 18: Is it possible to not use subpixels? If I recall correctly browsers render subpixels differently.

If there is no way around this, then leave it as is.

michaelpande (approved):

packages/pxweb2-ui/src/lib/components/Tag/Tag.module.scss
Glenruben: Discussed with designer, we agreed to round up to 2 pixels.

PerIngeVaaje (approved):

SjurSutterudSagen (approved):
LGTM

PR #651, opened 2025-04-25
Update project minor/patch version

SjurSutterudSagen:
Since we want to keep the project updated, we should update the project before releases.


PerIngeVaaje (approved):
OK

PR #655, opened 2025-04-29
Show status in datacells if provided

PerIngeVaaje:
Datacells with numbers returned from API could have a status connected. If so the status should be shown together or instead of the number


MikaelNordberg (approved):
Nice work! Looks good

PR #657, opened 2025-04-29
Add language to url for language persistance

SjurSutterudSagen:
Since we need to support several languages in the project, we need to add language persistence. What complicates this is that a table can be available in only one language, even though several can be supported for the entire site. And we also need it to work when people share links with other people.

This adds an optional language part to the url, and it uses the fallback language if that part is missing. This means that urls like '/sv/table/tableid' and '/table/tableid' both will work. Since it seems unlikely a site will change its default/fallback language.

I installed the i18next package for detecting browser languages, but set it up to only check for path, and to not cache the values. Since we only want the language in the URL to matter, and to avoid any situations where a user has an invalid language selected for the site, for the table they are trying to view.

Added tests, and also had to refactor some for the ErrorPage and ErrorBoundry. This work on the tests was done in large part with Claude 3.7.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6633,6 +6633,15 @@
         }
       }
     },
+    "node_modules/i18next-browser-languagedetector": {
+      "version": "8.0.5",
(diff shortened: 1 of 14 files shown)


MikaelNordberg (approved):
Looks good

PR #667, opened 2025-05-13
Fix missing translation keys in non-Eng languages

SjurSutterudSagen:
This adds the missing formatting translation keys for the sv, no and ar languages. This was not needed before, because we defaulted to the en language. It is needed now, since we want to add the ability to change which language is the default language.

This is needed to change the default languages in the two test sites.


PerIngeVaaje (approved):
OK

PR #669, opened 2025-05-14
Update variable mandatory status according to selected codelist

MikaelNordberg:
When a codelist is selected for a variable, the mandatory status for the variable shall be updated to reflect the fact that the codelist can be eliminated or not.


PerIngeVaaje (approved):
Lovely

PR #671, opened 2025-05-15
Wait until data is reloaded for fading

michaelpande:


michaelpande:
**Testing if you want:**
1. First load the table
2. Use a slow network (dev tools select 3G)
3. Add an additional month or another variable that was previously not in the set (so that we trigger a new request)

PerIngeVaaje (commented):
Looks like the table sometimes fades when it should not. 
First times loading

ssb-cgn (approved):
👍

PR #678, opened 2025-05-19
Fixed bug when checking if there are any changes in selected variables and values

MikaelNordberg:
In some cases changes in selected variables and was not detected.
There was a bug in the hasChanges function in the VariablesProvider. Instead of returning true from the function when there was a change in selection, the code just returned from the forEach-loop within the function. This is now fixed.


SjurSutterudSagen (approved):
LGTM

PR #688, opened 2025-05-21
Feature/px web2 432 add support for always default lang in url

SjurSutterudSagen:
This expands on the logic for handling persistence of language in the URL. 

1 - It generates up the specific routes needed for each supported language, which are defined in the config file. This also means unsupported languages can get an error message. Though there might be edge cases here I haven't encountered or thought about here.
2 - It adheres to the config option about showing the default language in the URL or not.
3 - Added a small NotFound component, for showing unsupported language and table not found. (Temporary, added to not have those routes show an empty page with only 1 line of text)

Also added tests.


SjurSutterudSagen:
I get an error in sonarqubecloud when I click the "1 New Issue". Can't see what is wrong there.

MikaelNordberg (approved):
Looks good

PR #699, opened 2025-05-28
Add ContentBox component

SjurSutterudSagen:
This adds the contentbox component to the library, which we can use to wrap the different operations we are going to have in the sidebar in the web app.

I am a bit unsure if it should be added to the web app components, or be added in the ui library like I have done.

Also added some basic tests and stories, and updated the NavigationDrawer to better fit the intended use.


PerIngeVaaje (approved):
Looks good!

PR #701, opened 2025-05-30
Load notes when changing codelist (pxweb2 652)

MikaelNordberg:
When changing codelist we need to load the notes for the new codelist.
We do this by calling the metadata-endpoint of the API in the context of the selected codelist.
Regenerated the API-client to the latest version of the specification.


PerIngeVaaje (approved):
Lovely!

PR #712, opened 2025-06-04
Add proper title handling to NavigationDrawer

SjurSutterudSagen:
Since we are starting to work on the different pages/views of the sidebar/NavigationDrawer component, we need to fix how it currently always displays the title for the filter view. Even on other views.

This adds the functionality for handling that. It needed some renaming for ease of use in the future, and to bring the NavDrawer closer to the translation keys.


SjurSutterudSagen:
> Hm, looks like there is an issue with this and the AccessiblityProvider, when it tries to start opened in DesktopMode. In the branch it no fails on an error:
> 
> `Cannot read properties of undefined (reading 'setAttribute')`

I had overlooked some types and names in the TableView page.

oyessb (approved):
Looks good 👍

PR #714, opened 2025-06-04
Add pivot clockwise to navigation drawer edit pane

SjurSutterudSagen:
This adds a button to pivot the table head and stub. Reused the function that was made some time ago. The pivot function itself has not been updated. This also means we need to add the dynamic title to ContentTop, which is based on the order of the variables in the stub and heading.

As part of this we noticed some performance issue, and several components and functions have been memoized. Additionally we saw that a major bottleneck was the formatting that was done on the values in the cells, and this formatting has been moved to the fetchFromApi function, which means it only happens once; when a value is loaded into the cube. Used this guide to using react-i18next outside of React to do the formatting: https://www.locize.com/blog/how-to-use-i18next-t-outside-react-components

During testing now, it seems like a deployed version uses about 1.5s to pivot a table with over 130,000 cells.

Also added support for 6 decimals. But we might want to have another look at reducing the number of formatting strings later, maybe we could use plurals for them?


MikaelNordberg (approved):
Looks good.
A possible improvement could be to set the formatted data cell value already in the mapping from jsonstat2.

PR #715, opened 2025-06-05
Fix keyboard navigation for single value in variablebox

oyessb:
Set dynamic tabIndex for Checkbox based on value count in VariableBoxContent


oyessb:
Not longer needed as we are going to use tab for navigation inside list. Closing PR.

PR #739, opened 2025-06-17
Use custom request.ts file that works for binary formats

MikaelNordberg:
Use custom request.ts file that works for binary formats, instead of using generated request.ts file.

When adding "--request ./packages/pxweb2-api-client/custom/request.ts" to the generate api-client script, no request.ts file will be generated. Instead this custom file will be used and placed in the folder./packages/pxweb2-api-client/src/core folder


oyessb (approved):
Looks good 👍

PR #755, opened 2025-06-26
Pxweb2 599 select with search

ssb-cgn:


packages/pxweb2-ui/src/lib/components/SearchSelect/SearchSelect.tsx
```diff
@@ -0,0 +1,247 @@
...
+              size="small"
+              onClick={handleClear}
+              aria-label="Fjern valg"
+            />
+          ) : (
+            <Icon
+              iconName="ChevronDown"
+              className={cl(styles.chevron, { [styles.open]: isOpen })}
+            />
+          )}
```
Glenruben, line 202: Nothing happens when I click this button. Maybe it needs an onClick event listener to open the select box!

Glenruben (approved):
Nice fix with pointer-events! Clean and tidy!

PR #756, opened 2025-06-27
Fix the installation of the i18next packages

SjurSutterudSagen:
The current installation of the different packages for i18next, are a bit inconsistent and missing. Some are installed in the web app, another is installed in the ui lib, and the i18next main package is missing entirely. This fixes the packages installations by installing them in the root package.json, so they can be used in the web app and ui lib as needed.

We can move them into specific packages later if we need to.


oyessb (approved):
Looks good 👍

PR #763, opened 2025-07-02
Fix import for List component in index.ts

SjurSutterudSagen:
Add import for the List component in the UI library's index.ts file, since it was missing.

Also remove a duplicate line in there.


PerIngeVaaje (approved):
OK

PR #768, opened 2025-07-07
Add support for translations in Storybook/UI-lib

SjurSutterudSagen:
This adds a first iteration of language support to the Storybook in the PxWeb2 UI library. The solution was made following Storybooks example for this, but I also made considerations for speed and reducing complexity.

Therefore, there are improvements that can be made to this. For instance, we could consider moving the translation into their own package, and import them into other packages as needed. Now it just imports directly from the web app.


PerIngeVaaje (approved):
OK

PR #769, opened 2025-07-07
Remove intermediate status from mixed checkbox

SjurSutterudSagen:
This removes the intermediate, visual style for the mixed status of the mixed checkbox. It retains the functionality, but now looks the same as the old uncheck/no values checked status. This also removes the different text values for "Select All/Deselect All", since those were found to be confusing with how this checkbox works during testing. Also cleanup relevant translations and translation logic.

Also expanded the tests for the Checkbox and MixedCheckbox components.


KentMossback (approved):
Looks good!

PR #773, opened 2025-07-11
Add first value when changing codelist for mandvar

SjurSutterudSagen:
This adds the logic for adding default values, when changing codelists for mandatory variables. The values chosen is limited to the first value, but the code should be easily extendable to be expanded later. This is why I chose to have the logic for adding the defaults seperated from adding codelist to variables. Any extra processing should not have a significant impact, since we are only mapping over variables here, and not values.

Also updated the tests.


PerIngeVaaje (approved):
OK

PR #775, opened 2025-07-11
Fix eslint warnings about PascalCase for Stories

SjurSutterudSagen:
This fixes all the current warnings about PascalCase for Stories, that we get when running the lint command. Many had camelCase, including one story file.


KentMossback (approved):
Looks good!

PR #784, opened 2025-08-07
Get saved query

PerIngeVaaje:
Load and display saved query in PxWeb.

If language is changed in PxWeb default selection is loaded in the new language.

Removed language in dependency array for fetching data. Now data fetch is triggered by change in tableId or variableSelection. Before API call to fetch data was called twice. Could cause timing problems depending on which call was ready first. We had a problem where selected variables and values in variable boxes differed from what was displayed in the table. This should now be fixed.


SjurSutterudSagen (dismissed):
We can discuss more about it later, but approving it now.

SjurSutterudSagen (approved):

PR #802, opened 2025-08-18
Change the PivotButton to have proper design

SjurSutterudSagen:
This changes the PivotButton to have the correct design and aria functionality. Refactored the ActionItem slightly, so it can have different label and aria-label.


MikaelNordberg (approved):
Looks good

PR #804, opened 2025-08-18
Remove contentcode variables from the table title

SjurSutterudSagen:
This removes the ContentCode variables from the table title, and also adds functionality for when there is only one variable for the table.

Still need to test this on a table with only one variable.


SjurSutterudSagen:
Closing this for now. Thought it would be 1 hour of work or thereabouts, but several issues with edge cases. Wrote more details in the JIRA issue in the backlog.

PR #817, opened 2025-08-22
Feature/pxweb2 807 hide drawer button

MikaelNordberg:
Hide drawer is now coded as a <button> instead of a clickable <div>


oyessb (approved):
Looks good 👍

PR #821, opened 2025-08-25
Fix/mixed checkbox still shows mixed state with keyboard

SjurSutterudSagen:
This changes the checked state of the MixedCheckbox component to only show:
 - "true" if all are selected
 - "false" if none or some values are selected

We need to know all three states to properly handle checking and unchecking, based on whether some/none or all are selected, so that logic was not change. This only changes the checked state, for screenreader users.

Also adds test to cover the aria checked state.


PerIngeVaaje (approved):
OK

PR #822, opened 2025-08-25
Feature/pxweb2 806 search contrast

MikaelNordberg:
To get better contrast in the Search component a border is added at the bottom.
Border radius has also been changed.


PerIngeVaaje (approved):
OK

PR #829, opened 2025-08-27
Breadcrumbs for navigation 

KentMossback:
Breadcrumbs component that let the user see where in the subject structure a table belongs. In the component it is possible to navigate back to the startpage.

Possible improvement for the feature is to use useNavigate, to reduce the number of server calls.


MikaelNordberg (approved):
Looks good

PR #831, opened 2025-08-28
Feature/px web2 135 header with language switcher select

SjurSutterudSagen:
This implements the intended design and functionality for the Header component, including implementing the design for the LanguageSwitcher.

The language switcher is made mostly with HTML and CSS, but had to add a check for keyboard navigation due to the HTML select's automatic re-focus when making any selection. Since this would show the focusmarking, even when making a choice by clicking with a mouse.

Also, this fixes the type and CSS for the Button's iconPosition. It was implemented with "left/right", however since we support languages with RTL writing, this does not make semantic sense. Change "left/right" to "start/end". All functionality of Button is the same, this is only the names.


PerIngeVaaje (approved):
OK

PR #851, opened 2025-09-08
Fix/test output spam

SjurSutterudSagen:
Remove console.log from startPageFilters test and fix typo

Wrap imperative ref call in act and replace setTimeout waits with waitFor in Search test


oyessb (approved):
👍

PR #863, opened 2025-09-10
Creates a package with only the contents of the dist folder

Glenruben:
the zip command is sensitive to the relative path of the files included, and needs to be run from the correct directory in order to not include the folder hierarchy in the archive contents.


MikaelNordberg (approved):
Looks good

PR #867, opened 2025-09-12
Footer component that you can configure

MikaelNordberg:
Footer component that can be configured using a configuration file per language.


SjurSutterudSagen (approved):
LGTM

PR #871, opened 2025-09-15
Patch/update minor and patch

SjurSutterudSagen:
Updated Storybook with the SB Cli, no automigrations needed.

Bumped the minor and patch versions in the project.

Did not install the jsdom major version.


oyessb (approved):
👍

PR #880, opened 2025-09-16
Changed `variable.css` so that it references other css variables 

likp:
Changed `outputReferences = true` option to CSS variable files so that it generates css variable references instead of the actual value.

### DIFF
--- a/packages/pxweb2-ui/style-dictionary/build.mjs
+++ b/packages/pxweb2-ui/style-dictionary/build.mjs
@@ -25,6 +25,7 @@ const configs = [
           {
             destination: 'variables.css',
             format: 'css/variables',
+            options: { outputReferences: true },
           },
         ],
       },
@@ -36,6 +37,7 @@ const configs = [
           {
             destination: 'variables.css',
             format: 'css/variables',
+            options: { outputReferences: true },
           },
         ],
       },


runejo:
@likp should we include the changed `variables.css`?  I see its on the .gitignore list

runejo:
> @runejo I am uncertain my self. We generate the `variables.css` and we do it twice. Once in `pxweb2-ui` and and one in the `pxweb2` under `public/theme`. I see that we have a ignore rule for `theme/variables.css` but since the `theme` folder is not directly in the root the `variables.css` in `pxweb2` is not ignored. The other `variables.css` in `pxweb2-ui` is ignored since it is place in the `dist` folder which has it´s on ignore rule.
> 
> My guess is that they are ignored since they are automatically generated so that we might as well ignore them. If so we should fix the ignore rule to be `**/theme/variables.css`.

This PR would be faster to review if it was not ignorede. The `main` branch has an up to date `theme/variable.css`, maybe this is whats making it confusing.

PerIngeVaaje (dismissed):
OK

likp:
> > @runejo I am uncertain my self. We generate the `variables.css` and we do it twice. Once in `pxweb2-ui` and and one in the `pxweb2` under `public/theme`. I see that we have a ignore rule for `theme/variables.css` but since the `theme` folder is not directly in the root the `variables.css` in `pxweb2` is not ignored. The other `variables.css` in `pxweb2-ui` is ignored since it is place in the `dist` folder which has it´s on ignore rule.
> > My guess is that they are ignored since they are automatically generated so that we might as well ignore them. If so we should fix the ignore rule to be `**/theme/variables.css`.
> 
> This PR would be faster to review if it was not ignorede. The `main` branch has an up to date `theme/variable.css`, maybe this is whats making it confusing.

Oh, I see. I add the the updated `variables.css` file.

runejo (approved):

PR #884, opened 2025-09-18
Remove unused forwardRef from Breadcrumbs

SjurSutterudSagen:
This removes an unused forwardRef on the Breadcrumbs component, which was spamming output into every test during test runs. Also, we should not add forwardRef anymore, since it is no longer needed.

Also added some empty lines to divide up the code a bit, and made the props readonly.


PerIngeVaaje (approved):
OK

PR #916, opened 2025-10-03
Feature/simplify showing correct error messages

SjurSutterudSagen:
The error handling in 8e08e51f42d155190f4cd145237e0abcd129214d was overengineered, this implements all the various simplifications of the older handling, while also adding a simpler solution to the problem fixed in that squashed PR.


ssb-cgn (approved):
👍

PR #942, opened 2025-10-20
Fix/LanguageSwitcher uses wrong language for checks

SjurSutterudSagen:
This fixes a bug where the LanguageSwitcher was using the fallback language in the config, when it should have used the default language. We did not find this bug sooner, since we have always ran the web app with the same fallback and default languages.


MikaelNordberg (approved):
Looks good

PR #943, opened 2025-10-21
Fix/i18next should use default lang

SjurSutterudSagen:
This fixes the logic for the language system, so it uses the default language when no language match is found in the URL. It is needed when the default language should not be in the URL.

It does this by adding a new, custom detector that always returns the default language, from the config file. This is put behind the URL matching, so it will only be used when no language matches.

It is reliant on https://github.com/PxTools/PxWeb2/pull/942 to fully fix the language issues in https://statistics-norway.atlassian.net/browse/PXWEB2-894


MikaelNordberg (approved):
Looks good

PR #1196, opened 2026-03-30
Omit empty target attribute when not provided

oyessb:
html-validate reported invalid emptry string for target attribute


MikaelNordberg (approved):
Looks good

PR #1223, opened 2026-04-15
feat: add .npmrc for npm configuration and document dependency install cooldown

SjurSutterudSagen:
The new npmrc file allows us to set NPM options in the project, instead of having to remember to adding the relevant flags to commands.

This should improve both developer workflow and security.


MikaelNordberg (approved):
Looks good

PR #1226, opened 2026-04-16
patch: Upgrade project dependencies 20260416

SjurSutterudSagen:
Tested upgrading with a local .npmrc file with:
min-release-age=9
ignore-scripts=true

It seem to work great, both with the storybook CLI (i used @(version), not @latest), and the other minor and patch version. At least when i did it in that order.


MikaelNordberg (approved):
Looks good

PR #1228, opened 2026-04-20
chore: Upgrade i18next major versions

SjurSutterudSagen:
Upgrades i18next and react-i18next to their latest major versions.

"count" now seems to be a protected keyword in, so had to refactor a bit where this led to a typescript error.

Also fixed another typescript error in the StartPage.tsx file.


MikaelNordberg (approved):
Looks good

PR #1230, opened 2026-04-20
chore: update dependencies and configure babel plugin for Vite

MikaelNordberg:
- Added @vitejs/plugin-react to the main package.json and updated its version.
- Removed old @vitejs/plugin-react dependency from pxweb2-ui package.json.
- Removed babel-plugin-react-compiler from pxweb2 package.json.
- Updated vite.config.ts in pxweb2 to use @rolldown/plugin-babel with reactCompilerPreset.


SjurSutterudSagen (dismissed):
Looks good, just double check that it is the correct color code before merging

SjurSutterudSagen (approved):

PR #1236, opened 2026-04-22
Fix html validation error table page. Rule element-permitted-content

PerIngeVaaje:


package-lock.json
runejo, line 1: Why is the lock file changing?

PerIngeVaaje: It happened after pullling  newest main and new install

runejo: I think maybe not all developers have upgraded to Node v24 :-)

SjurSutterudSagen: Some had an older NPM version in the path. :)

MikaelNordberg (approved):
Looks good

PR #1250, opened 2026-04-29
Fix wrong use of label inside button

PerIngeVaaje:
Change div to button and use Label as span element


SjurSutterudSagen (approved):
LGTM, seems like a better solution this one

PR #1252, opened 2026-04-30
feat: Add parsing definitions from the api calls

SjurSutterudSagen:
Add parsing for the definitions in the metadata API call into the jsonStat2ResponseMapper, make the TableInformation component use that for the DefinitionsTab.

This will make the DefinitionsTab show up again, if the API has the correct definitions in the response.


MikaelNordberg (approved):
Looks good

PR #1259, opened 2026-05-07
feat: Create badge component in UI library

SjurSutterudSagen:
Creates the Badge component in the UI library, also refactors the Tag components props to match how it is done in the Badge component.


MikaelNordberg (approved):
Looks good

Issue #1229, opened 2026-04-20
Github actions with deprecated node 20 version warnings


SjurSutterudSagen:
We currently have 3 actions that use a deprecated version of Node (20) that will be affected when Github removes support for that versions in their runners on June 2. https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

1. powerful-pr-comment: The action for setting the comment on PRs with the 4 different PR links, for testing. This one seems not to be maintained, since it has no activity in the last 2 years. We should look for an alternative here. https://github.com/jupier/powerful-pr-comment
2. github-run-tests-action: This is the mabl action, and there is an approved PR there with the update to node 24. Just not merge yet. https://github.com/mablhq/github-run-tests-action/pull/69
3. cloudflare/wrangler-action: The cloudflare deploy action, also has a PR with the node 24 support. https://github.com/cloudflare/wrangler-action/pull/412

SjurSutterudSagen:
The last action had a new release with the fix 14 hours ago

SjurSutterudSagen:
We updated the last action here: https://github.com/PxTools/PxWeb2/pull/1348
These are all fixed now, closing this issue.

PR #1276, opened 2026-05-20
chore: Update wrangler-action to version 4.0.0 in deployment workflow

SjurSutterudSagen:
Updates the action to a version that supports node 24

Fixes number 3 in https://github.com/PxTools/PxWeb2/issues/1229


MikaelNordberg (approved):
Looks good

PR #1285, opened 2026-05-26
chore: Pin most actions to commit SHAs

SjurSutterudSagen:
This pins most of the actions to commit SHAs, instead of version tags. This hardens the project against attacks that change which commit a tag is pointing to for the actions. However, there are a couple of the actions left that were problematic:

1. github/codeql-action: I haven't found where to find the correct SHA for the releases of the "Upload sarif" action
3. cloudflare/wrangler-action: Seems to have a weird setup for their releases where the SHA is not in that repo? Has this warning: "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

Also, we should look at the settings for the repo, and consider forcing pinning actions to SHAs in there too. There is an option if you have enough access, to force it for actions.

### DIFF
--- a/.github/workflows/chromatic-skip.yml
+++ b/.github/workflows/chromatic-skip.yml
@@ -22,12 +22,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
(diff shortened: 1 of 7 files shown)


SjurSutterudSagen:
There is currently problems with github actions worldwide: https://www.githubstatus.com/

We can see if any actions have started failing after they are back to normal

MikaelNordberg (approved):
Looks good

PR #1286, opened 2026-05-27
chore: Cleanup some spam in tests

SjurSutterudSagen:
The output from the tests have started to spam with different warnings etc. This PR cleans up some of it, but there are a few places left that still needs more cleanup work.

I figured it's better to get this inn sooner rather than later.


MikaelNordberg (approved):
Looks good

PR #1288, opened 2026-06-01
Use pinact tool to pin GHA versions to SHAs

SjurSutterudSagen:
Pinact supports flags for min-age and verify-comment

Ran command `pinact run -min-age 9 -verify-comment`, which mostly added a blank space in the version comments. But also handled the two problematic actions

Nor sure if the devs who use Windows can use it, but works great for us on Macs


MikaelNordberg (approved):
Looks good

PR #1307, opened 2026-06-05
fix: ensure parent subject filter is correctly restored during sibling removal

### LINKED ISSUE #1289: Sometimes the filter on the start page show wrong chips
### Describe the bug
When you add 2 children to a filter, and then remove one of them, the filtering on the startpage counts both the single child and the parent as active. This is not correct, and shows too many chips for instance.

### Steps to reproduce

1. Choose a filter
2. Choose 2 new filters under the previous filter
3. Remove 1 of the 2 filters

Image

SjurSutterudSagen:
This fixes a bug in the "subjects" filter, where the parent is added when a child is removed, even if there still are other siblings still active

Fixes https://github.com/PxTools/PxWeb2/issues/1289


MikaelNordberg (approved):
Looks good

PR #1308, opened 2026-06-05
fix: Make the variable filter's height dynamic

SjurSutterudSagen:
This makes the height of the Variablefilter on the startpage dynamic to the content.


SjurSutterudSagen:
@Maren-SSB Liked this over the old way, so we can review and merge it

MikaelNordberg (approved):
Looks good

PR #1309, opened 2026-06-08
chore: Update concurrently from v9 to v10

SjurSutterudSagen:
Updates the major version of the concurrently package. Tested locally with both version, and runs as expected with the same output.


MikaelNordberg (approved):
Looks good

PR #1315, opened 2026-06-09
build(deps): update Node.js version from 24.14.1-slim to 24.16.0-slim…

SjurSutterudSagen:
… in Dockerfile

I think we have gotten dependabot PRs for the new 26.x versions, but they don't enter LTS until the end of the year, so this one was missed.


MikaelNordberg (approved):
Looks good

PR #1316, opened 2026-06-09
fix(coverage): include additional packages in coverage exclusion list

SjurSutterudSagen:
This removes the duplicate entries in the output when running the coverage command from the root. 

The web app's coverage command now only contains its own coverage output.


MikaelNordberg (approved):
Looks good

PR #1320, opened 2026-06-09
chore: Cleanup remaining test output spam

SjurSutterudSagen:
- Moved api-client mock, and make it automatic for all tests
- Remove a console.error (that also throws) that is not needed, but shows up in test output
- Cleanup mock data structures in the JsonStat2mapper


MikaelNordberg (approved):
Looks good

PR #1321, opened 2026-06-09
chore: Update project minor and patch versions

SjurSutterudSagen:
Updates the projects minor and patch versions. No storybook updates so using their CLI was not needed.


MikaelNordberg (approved):
Looks good

PR #1322, opened 2026-06-10
fix: Override select width in ApiQuery

SjurSutterudSagen:
Adds an override for the width of the Select in the ApiQuery, making it take 100% of available space.

Also noticed the ApiQuery component did not have any unit tests, so used AI to generate some.


MikaelNordberg (approved):
Looks good

PR #1324, opened 2026-06-11
chore: Update workflow to include Node.js version 26.x for builds

SjurSutterudSagen:
This is so we can start seeing any warnings etc. with the newer node and npm versions. They have announced some changes that might affect us.

Also found an old node 22 version that seems to have fallen through the cracks.


SjurSutterudSagen:
In the `Use Node.js 26.x` action it says `npm: 11.16.0`, so used the correct npm version for the warnings. And didn't find any.

runejo (approved):

PR #1336, opened 2026-06-19
chore: Migrate from vite-plugin-dts to unplugin-dts

SjurSutterudSagen:
This swaps the old package "vite-plugin-dts" for the new package "unplugin-dts". They added support for more than just Vite, which is why they changed the name. They recommend using the new package, instead of continuing using the old one, even though they still update it (it is now just a wrapper that reexports from the new one).

- Added "unplugin-dts" to package.json for improved TypeScript support.
- Updated vite.config.ts in pxweb2-api-client and pxweb2-ui to use "unplugin-dts" instead of "vite-plugin-dts".
- Removed "vite-plugin-dts" from pxweb2-ui package.json as part of the migration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,6 +47,7 @@
         "jsdom": "^29.1.1",
         "style-dictionary": "^5.4.4",
         "typescript": "^6.0.3",
+        "unplugin-dts": "^1.0.2",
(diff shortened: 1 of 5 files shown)


PerIngeVaaje (approved):
OK

PR #1337, opened 2026-06-19
chore: update html-validate to version 11.5.3 and adjust configuration

SjurSutterudSagen:
The "void" rule was removed quite some time ago, but it seems we had not updated it. Asked AI and the "void-style" fits the old rule.

Tested before and after, and seems to give the same output.


MikaelNordberg (approved):
Looks good

PR #1346, opened 2026-06-23
feat: Chart autopivot

### LINKED ISSUE #176: Graph - Automatically show the best chart visualisation
Automatically show the best chart visualisation

When users select the line chart, they should automatically be given the best visualisation possible:
Time should be preferred as category axis if >1 time is selected

**Note**
The "Improve table layout" button is not modified to also work in chart mode. We decided to create a issue of its own for this. https://github.com/PxTools/pxtools-project/issues/184

**Acceptance criteria**
When selecting the line chart you automatically get the best visualisation (based on rules) 

**Test**
Verify that chart is displayed in the expected way with time on the x-axis.
Also verify that table (both desktop and mobile) still is displayed in the expected way. 
Verify that pivoting works as expected. Switch between the different presentation views (desktop-/mobile table and chart) and verify that pivoting works as expected in all of the views.
Change the selection. Add and remove variables and values. Change codelists. Switch between presentation views and verify that it works as expected. 
**Note! Charts can only be tested using the swedish test links since charts are still not enabled in the norwegian configuration file.**
**Question: What shall happen if we have no time variable?**

MikaelNordberg:
When chart view is selected data is automatically pivoted to display the chart in the best way possible.

TableDataProvider now handles DataViewMode instead of isMobile. In this way data is pivoted for best view also for charts and not only mobile/desktop table.


SjurSutterudSagen:
The code looks good to me now. Due to vacations I won't be able to approve it after the testers have looked at it.

PerIngeVaaje (approved):
Good!

PR #1351, opened 2026-06-29
chore: Upgrade react-router from v7 to v8

SjurSutterudSagen:
This updates to the new, v8, major version of React Router. There were no breaking changes that affected us, but we fixed a minor import issue in main.tsx. And now we import the RouterProvider from the correct sub-package.

Co-Author: @MikaelNordberg


MikaelNordberg (approved):
Looks good

PR #1358, opened 2026-07-02
feat: Improve chart axis functionality

### LINKED ISSUE #113: Graph - Display unit/variable name at x-axis and y-axis
The axis that shows figures (content variable) should display measurement unit for the variable at the end of the axis. This requires a control that not multiple units are selected.

The other axis (categories) should display variable text at the end of the axis.

This issue must also include @BjarteMell's comments about line chart: https://github.com/PxTools/pxtools-project/issues/18#issuecomment-4726554703

**Test**
Verify that y-axis is broken when needed as described in Bjartes comments.
Verify that unit and variable name is displayed on the y- and x-axis.
Verify that a message "cannot display chart" is displayed when content with multiple units (for example number and percent) is selected 
Tests should be made via the swedish test links.

MikaelNordberg:
Display measurement unit for the variable at the end of the axis.
Adaptive tick marks on y-axis.
Do not show chart when data contains contents with different units.


SjurSutterudSagen (approved):
Good job, great to see the clean up of the Presentation logic too!

PR #1360, opened 2026-07-06
chore: Add release-please configuration and workflow files

### LINKED ISSUE #114: Make immutable releases for PxWeb2
**Description**
We should increase the hardening in PxWeb2 releases by making new releases “Immutable” (e.g. that no one can replace the zip file in eh releases after it has been published).

It seems that GitHub has added support for this, but our create-release.yml workflow will not work with this. We first create the release, and then we build the artifacts/zip file that is added to the release afterwards. And this must then be done before the release is published.

Immutable releases - GitHub Docs See also “Best practices” directly below the link.

The background to the task is that as of today (March 25, 2026) there is an ongoing attack on Trivy, where an attacker has replaced assets (zip file etc. on all release/version tags) in their repo. It seems that this will not be stopped by our current branch protection setup, as it does not touch main (only the actual commits the tags point to.

Trivy Security incident 2026-03-19 · aquasecurity trivy · Discussion #10425

**Acceptance criteria**

Immutable releases are turned on for releases in PxWeb2
New releases are created with correct assets and artifacts.

### LINKED ISSUE #134: Improve clarity and structure of release notes
_Originally opened by Kristin in Jira, March 2026._

Det blir flere og flere brukere/organisasjoner som følger med på PxWeb 2 sine release-notes, og fint om de kan grupperes etter endringstype, så de blir enklere å forstå for de som følger med fra sidelinjen. 

Mimir har et godt forslag på hvordan en release-notes kan være: https://github.com/statisticsnorway/mimir/releases

runejo:


SjurSutterudSagen (dismissed):
Great to finally land this! Approved!

SjurSutterudSagen (dismissed):

SjurSutterudSagen (approved):

PR #1369, opened 2026-07-07
feat: show PxWeb version

### LINKED ISSUE #101: Show PxWeb-version in the interface
As a user of PxWeb in my organization, I want to be able to find the version number to be able to communicate it if I file an issue on Github for example.

From API meeting 2026-01-20
Version number in PxWeb
Modify the release pipeline so that embed the tag into index.html or in a separate file.
SCB and SSB need to replicate this to our own pipeline.

**Acceptance criteria**
A user can find the release version displayed hidden in the interface

runejo:
Instead of `BUILD_DATE` we now use `APP_VERSION` to do cache busting. This can also be used to verify the version of PxWeb by using "view source" in a browser


SjurSutterudSagen (approved):
LGTM, this might also get rid of some build warnings about __BUILD_DATE__ only being known at runtime.

Just checked and those warnings are removed!

PR #1420, opened 2026-08-11
fix: Wrap the search in FilterSidebar in useMemo

SjurSutterudSagen:
Wraps the debouncedSearch in FilterSidebar with useMemo. The old code added the debounce inline, which makes it potentially get recreated on every rerender of the component.


PerIngeVaaje (approved):
OK

PR #1433, opened 2026-08-12
build: Add "nodejs_compat" flag to wrangler configuration

SjurSutterudSagen:
We see that the cloudflare deploy on main fails with the wrong Node version, even though the dashboard has the NODE_VERSION environment variable set to 24. This adds the "nodejs_compat" compatability flag to the wrangler.jsonc file. It might change which nodejs version is chosen based on the compatability_date, which is also updated to today.


PerIngeVaaje (approved):
OK

PR #1434, opened 2026-08-12
build: Add vars NODE_VERSION to wrangler.jsonc

SjurSutterudSagen:
The configurations set in the Cloudflare project does not seem to be read, and logs show the deploy on main find no environment variables.

This adds a vars object in the wrangler.jsonc file, with a NODE_VERSION set. Which will hopefully set this version in the cloudflare runs.

If this does not work, we might have to look into doing the Cloudflare Pages deploy of main like the ones in PRs.


PerIngeVaaje (approved):
OK

PR #1436, opened 2026-08-12
chore: Add missing engine.npm version to pkg-lock

SjurSutterudSagen:
When we added the enforcement of the Node and NPM versions, we forgot to run 'npm install' to also update those parts of the package-lock file.

This commit adds that missing part. I ran 'npm install' only.


PerIngeVaaje (approved):
OK

PR #1442, opened 2026-08-17
fix: definitionstab show only statistics homepage if wanted

KentMossback:
This fix ensures that the definitions tab is rendered correctly when data is fetched from the API metadata endpoint.

In the current version of PxWeb2, the Definitions tab is only rendered if data for "about-statistics" is present. This causes a problem when only "statistics-homepage" information is available from the api, as is the case for Sweden.

With this fix, the definition tab will be rendered if eihter "about-statistics" or/and "statistics-homepage" data is present for a table. Relevant LinkCard components will be shown.


SjurSutterudSagen:
The title of the PR should follow conventional commit style. So should be "fix: " instead of "Fix/"

SjurSutterudSagen (approved):
Good change, better that either links can be used. More flexible this way. Approved!

PR #1459, opened 2026-08-25
fix: Resolve two type errors in Presentation component

SjurSutterudSagen:
The t-function expects strings for the values, while they are stored as numbers in the config file. This wraps the functions for getting the values in `String()`.


MikaelNordberg (approved):
Looks good

PR #1466, opened 2026-08-28
feat: chart tooltip improvements

### LINKED ISSUE #107: Graph - Tool tip
In the workshop we looked at two different ways of showing the graph tool tip: one value or all values. We need to explore and test these two different approaches to find the most user-friendly solution.

For more information about Apache eCharts tooltips: https://echarts.apache.org/en/option.html#tooltip
<br>

Developer, designer and Bjarte (?) work together to figure out the best solution.

## Test
In the test branch two different charts are displayed, one with a tooltip for all of the values, and a second chart that only shows the tooltip for one value at a time.

Since this task is a bit exploring, the testers test the different types of tooltips and gives feedback and then we develop further. And so on...

## Acceptance criteria

- Both tooltip approaches have been tested in realistic use cases.
- Edge cases (many values, long value names) have been considered.
- The chosen solution works across relevant screen sizes, graph types, with zoom etc.
- Follows WCAG success criteria at levels A and AA.

MikaelNordberg:
Tooltip text is only displayed for one data series at a time.
The tooltip remembers which series was last hovered and displays those values. When hovering another data series the tooltip switches to display that data series instead. 
On mobile you can also click to show tooltip.


PerIngeVaaje (approved):
OK

PR #1476, opened 2026-09-01
fix: Add correct styling to linechart showmore button

SjurSutterudSagen:
This adds the correct looks and functionality to the LegendToggleButton in LineChart. It does not touch the actual Legend area itself.


SjurSutterudSagen:
The "features: chartEnabled" will be removed before merging

packages/pxweb2-ui/src/lib/components/Chart/LineChart/LineChart.tsx
```diff
@@ -220,12 +223,27 @@ export function LineChart({
 interface LegendToggleButtonProps {
   readonly onClick: () => void;
   readonly text: string;
+  readonly isExpanded: boolean;
 }
 
-function LegendToggleButton({ onClick, text }: LegendToggleButtonProps) {
```
PerIngeVaaje, line 225: Should maybe be a separate component when we implement more graph types

PerIngeVaaje (approved):
Looks good

packages/pxweb2-ui/src/lib/components/Chart/LineChart/LineChart.tsx
SjurSutterudSagen: Indeed

PR #1480, opened 2026-09-03
fix: Eslint 10 compatibility

SjurSutterudSagen:
This cleans up our Eslint config, by removing unused lint-packages, and by removing the problematic eslint-plugin-react. That package still does not support Eslint 10, and Eslint 9 has been EOL for almost a month.

Lint output is the same as before, though we have lost some rules. We can reevalute the package when it supports Eslint 10. Or go with an alternative.

The biggest change for everyday development might be the removal of the unused imports rules/package.


PerIngeVaaje (approved):
OK

PR #1494, opened 2026-09-11
build: Exclude wrangler with deps from npmrc cooldown

SjurSutterudSagen:
This adds exclusions for Wrangler and its 2 dependencies to an exclusion array in .npmrc. 

I created this as an example on how to do it in our current setup, we should discuss if we want to merge this at all. The version installed here is 15 hours old as of now.

We would have to remove these exclusions later, so we don't always grab the latest Wrangler. And we could probably also say specific versions to grab in here, instead of grabbing any new version.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,6 @@
 min-release-age=9 # Enforce a dependency release cooldown of x days.
+min-release-age-exclude[]=wrangler
+min-release-age-exclude[]=miniflare
+min-release-age-exclude[]=workerd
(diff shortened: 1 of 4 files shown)


SjurSutterudSagen:
Not sure why those 5 optional deps went missing, but re-added them with "npx --yes npm@11.13.0 install --package-lock-only"

SjurSutterudSagen:
Optional dependencies seem to still be missing, stopping work on this now. The goal was not to actually do it, but to show roughly how to do it. Which this shows.

